PR #3, opened 2019-03-26
Create new azure pipeline for building reference-packages

dseefeld:


azure-pipelines/jobs/ci-linux.yml
```diff
@@ -0,0 +1,82 @@
...
+    stagingDirectory: $(rootDirectory)/source-build-reference-packages/staging
+
+  steps:
+  - template: ../steps/docker-cleanup-linux.yml
+
+  # Create working directory and copy source into it.
+  - script: |
+      $(docker.run) $(docker.root.map) $(docker.agentSrc.map) $(docker.agentSrc.work) $(imageName) bash -c '
+        mkdir -p /root/source-build-reference-packages/staging
+        rm -rf /root/source-build-reference-packages/
```
crummel: Isn't this deleting the directory we just created?  (and some other stuff)

dseefeld: That is definitely an error!  Thanks for catching it!

PR #4, opened 2019-04-01
Add packaging and build

dseefeld:
Add pipeline definition to:
* Build a tarball of the reference packages
* Build an empty project to get a parallel nupkg to publish
* Push tarball and nupkg to azure
* Generate a manifest
* Publish to Build Asset Registry


eng/Versions.props
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project>
+  <PropertyGroup>
+    <!-- This repo version -->
+    <MajorVersion>3</MajorVersion>
```
crummel: Starting at version 3.0 seems a bit odd - is this something we need to do for ref package versioning?

dseefeld: I was matching it to Release 3.0, but I see your point.  I can change it to 1.0.

PR #38, opened 2019-08-23
Update csproj files to build with correct keys

dseefeld:
Add keys for signing assemblies and update each project to specify the correct key to sign the assemblies.


src/referencePackages/src/microsoft.dotnet.platformabstractions/2.0.0/Microsoft.DotNet.PlatformAbstractions.csproj
```diff
@@ -3,6 +3,7 @@
   <PropertyGroup>
     <TargetFrameworks>netstandard1.3;net45</TargetFrameworks>
     <NuspecFile>$(ArtifactsBinDir)microsoft.dotnet.platformabstractions/2.0.0/microsoft.dotnet.platformabstractions.nuspec</NuspecFile>
+    <AssemblyOriginatorKeyFile>$(KeyFileDir)AspNetCore.snk</AssemblyOriginatorKeyFile>
```
crummel: It seems weird that PlatformAbstractions would have the ASP.NET key.  Are these auto-generated or did you do them manually?

dseefeld: I thought that one was weird, too.  They are auto generated.  I downloaded and double checked the assemblies in this package and it was correct.

PR #70, opened 2019-09-25
Refactor code to include targetPacks in one area

adaggarwal:


adaggarwal:
Ready for review.

adaggarwal:
> I don't understand why there are il src file changes. Shouldn't they be just straight moves?

I think everytime I'd run Ildasm, it changes small version/build numbers? And I ran again because the other step-1 tool in s-b-utilities PR also went through some cleanup and I generated the src again.

PR #78, opened 2019-10-10
Add sys tuple

adaggarwal:
Fixes part of [#1286](https://github.com/dotnet/source-build/issues/1286)


src/referencePackages/src/microsoft.build.tasks.core/15.1.1012/Microsoft.Build.Tasks.Core.csproj
```diff
@@ -73,6 +73,7 @@
         <PackageReference Include="Microsoft.Build.Utilities.Core" Version="[15.1.1012]" />
         <PackageReference Include="Microsoft.NETFramework.ReferenceAssemblies.net46" Version="1.0.0-alpha-5" />
         <Reference Include="System.Xml" />
+        <Reference Include="System.Xml" />
```
crummel: Duplicate `System.Xml` reference.

dseefeld: Weird.  I encountered this when I generated pkgs, too.  One time it generated and the next it didn't.  I added it in fixups because it was missing. :(

adaggarwal: Ya. That's true. For now I changed it manually though.

PR #103, opened 2020-02-07
Remove checked in binaries

dseefeld:
Download previously source-built artifacts rather than having them checked in to the repo.
Add code to suppress prevalent ilasm warning since there is no way to suppress warnings in ilasm.


src/dir.props
```diff
@@ -3,5 +3,9 @@
 
   <PropertyGroup>
     <ArtifactsRefPkgDir>$(RepoRoot)artifacts/reference-packages/</ArtifactsRefPkgDir>
+    <SourceBuiltDir>$(RepoRoot)artifacts/source-built/</SourceBuiltDir>
+    <SourceBuiltArtifactsTarballDownloadDir>$(RepoRoot)artifacts/downloads/</SourceBuiltArtifactsTarballDownloadDir>
+    <SourceBuiltArtifactsTarballName>Private.SourceBuilt.Artifacts</SourceBuiltArtifactsTarballName>
+    <SourceBuiltArtifactsTarballUrl>https://dotnetcli.azureedge.net/source-built-artifacts/assets/</SourceBuiltArtifactsTarballUrl>
```
adaggarwal: This is causing source-build bootstrap step 2 to fail. Because that step is offline and this would mean that we require online access. 
```
/bst/bootstrap_dir/reference-packages/src/setupSourceBuiltArtifacts.proj" (default target) (1) ->
       (DownloadSourceBuiltArtifacts target) -> 
         /bst/bootstrap_dir/reference-packages/src/setupSourceBuiltArtifacts.proj(12,5): error : MSB3923: Failed to download file "https://dotnetcli.azureedge.net/source-built-artifacts/assets/Private.SourceBuilt.Artifacts.0.1.0-3.1.100-preview3.dev.19571.3.tar.gz".  Name or service not known

```
Link - https://dev.azure.com/dnceng/public/_build/results?buildId=512771&view=logs&j=433fc19c-35e6-567f-bb68-1b20928cfb4e&t=c0d8a789-a37a-5a12-380a-c125bdeecb92&l=138 (the link may break after the build gets removed for AzDo)

PR #105, opened 2020-02-25
add msnetcoreapp 200

adaggarwal:


adaggarwal:
> I can't put the comments inline because the file is too large. What version of ildasm did you use to generate these? In System.Data.Common.il line 788 (and a few other lines), it uses the term "constraint" which is a keyword in the updated ilasm. The same issue exists in System.Reflection.Metadata.il line 23102. I'm not sure if this is an issue or not, but the versions of these files that were generated by source build had these keywords in single quotes (i.e 'constraint').

Thanks for catching that. The internal tool uses the 3.0.0.p9 whereas the one that source-build-reference-packages puts in artifacts is 3.1.0-rtm. I can update the tool to use the same version of core-clr tools.

dseefeld:
> > I can't put the comments inline because the file is too large. What version of ildasm did you use to generate these? In System.Data.Common.il line 788 (and a few other lines), it uses the term "constraint" which is a keyword in the updated ilasm. The same issue exists in System.Reflection.Metadata.il line 23102. I'm not sure if this is an issue or not, but the versions of these files that were generated by source build had these keywords in single quotes (i.e 'constraint').
> 
> Thanks for catching that. The internal tool uses the 3.0.0.p9 whereas the one that source-build-reference-packages puts in artifacts is 3.1.0-rtm. I can update the tool to use the same version of core-clr tools.

Sorry, I went back and looked at this in a 3.1 tarball and see that _constraint_ isn't treated as a keyword.  I must have been looking at a version that was generated with 5.0 tools.  I think this is good to go, then.

PR #113, opened 2020-05-08
Add microsoft.netcore.app.ref/3.1.0

dseefeld:


dagood:
Looks like this has the `System.Runtime.CompilerServices.NullableAttribute`s that had to be filtered out for the 3.0.0 targeting pack (https://github.com/dotnet/source-build/issues/1413), will that cause a problem?

dseefeld:
> Looks like this has the `System.Runtime.CompilerServices.NullableAttribute`s that had to be filtered out for the 3.0.0 targeting pack ([dotnet/source-build#1413](https://github.com/dotnet/source-build/issues/1413)), will that cause a problem?

Yes.  Thank you!  These two assemblies were not being produced and it was failing silently.  I've updated the ILSrc and the command to call ilasm to fail when necessary.

PR #118, opened 2020-05-28
Add NETStandard.Library.Ref,2.1.0

dseefeld:


dagood:
Does this need a `netstandard.library.ref.proj`? I don't see one in the diff, at least.

dseefeld:
> Does this need a `netstandard.library.ref.proj`? I don't see one in the diff, at least.

Yes.  I missed that.  Adding now.

PR #119, opened 2020-05-29
Add missing project for netstandard.library.ref,2.1.0

dseefeld:


dseefeld:
There were two projects I needed to add. :(

PR #121, opened 2020-06-06
Exclude obj directories from targeting packs.

crummel:
These don't exist in the official packages.


dagood:
Has this been added to the deconstructing tooling?

crummel:
Sorry, I'm not sure what you mean.  The `obj` directories come from building SBRP, not from  deconstruction+adding them.

crummel:
Oh, for generating the nuspecs.  Good point, I will open an PR for that.

PR #122, opened 2020-06-12
[dev/source-build-5.0] Make SBRP a more ordinary Arcade-using repo

dagood:
Initial work on https://github.com/dotnet/source-build/issues/1591 while keeping artifacts the same (modulo some incidental fixes: https://github.com/dotnet/source-build/issues/1619, https://github.com/dotnet/source-build/issues/1620). I didn't go quite as far as `ildasm`ing my outputs to compare vs. current outputs, but I diffed nupkg contents, and the DLL identity/version as reported by `FileVersionInfo`.

See commit messages for details. The last commit in particular has the build mechanism changes and probably the best to review. The others are generally mass renames.

No CI yet. I'm running with local `./build.sh` in a CentOS 7 build container (just for a little filesystem isolation while authoring MSBuild).


src/referencePackages/Directory.Build.props
```diff
@@ -0,0 +1,92 @@
...
+
+  <Import Project="$([MSBuild]::GetPathOfFileAbove(Directory.Build.props, $(MSBuildThisFileDirectory)..))" />
+
+  <PropertyGroup>
+    <!-- Common properties for all Reference Packages -->
+    <DebugType>None</DebugType>
+    <DebugSymbols>false</DebugSymbols>
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+
+    <AssemblyOriginatorKeyFile>$(RepoRoot)src/referencePackages/MSFT.snk</AssemblyOriginatorKeyFile>
```
dseefeld, line 25: This file doesn't exist?  Move below next line and use `$(KeyFileDir)`?  Or is this a catch in the case that the project doesn't define `AssemblyOriginatorKeyFile`?  If the latter, maybe a comment stating that?

src/referencePackages/src/dir.props
```diff
@@ -1,30 +1 @@
...
-    <AddAdditionalExplicitAssemblyReferences>false</AddAdditionalExplicitAssemblyReferences>
-    <GenerateTargetFrameworkAttribute>false</GenerateTargetFrameworkAttribute>
-    <RuntimeMetadataVersion>v4.0.30319</RuntimeMetadataVersion>
-
-    <EnableDefaultCompileItems>false</EnableDefaultCompileItems>
-    <GenerateAssemblyInfo>false</GenerateAssemblyInfo>
-    <WarningsAsErrors></WarningsAsErrors>
-  </PropertyGroup>
-</Project>
+<Project />
```
dseefeld, line 1: Is this file still needed?

src/referencePackages/Directory.Build.props
dagood: This is just moved from here, not sure why it exists--the file doesn't exist in master either:

https://github.com/dotnet/source-build-reference-packages/blob/aa0b7435c220ae8ef247e775f32eac3dbeb7162c/src/referencePackages/src/dir.props#L11

I'll try removing it.

src/referencePackages/src/dir.props
dagood: It's referenced in every csproj, keeping this file to avoid a mass edit for now. I'll add a comment.

dagood:
Merging without squash to preserve blame with this first set of mixed large-scale changes.

PR #123, opened 2020-06-15
Build intermediate nupkgs when /p:FromSource=true

dagood:
Bring forward prototype code from https://github.com/dotnet/source-build/issues/1543 and https://github.com/dotnet/source-build/issues/1466 to generate an intermediate nupkg when running `./build.sh /p:FromSource=true`.

The results of ordinary `./build.sh` are basically:

```
artifacts/packages/Debug/Shipping/{outputs}.nupkg
```

Then with `./build.sh /p:FromSource=true` you get an isolated artifacts dir and an output intermediate nupkg:

```
artifacts/source-build/self/packages/Debug/Shipping/{outputs}.nupkg
artifacts/packages/Debug/NonShipping/Microsoft.SourceBuild.Intermediate.source-build-reference-packages.linux-x64.5.0.0-dev.nupkg
```

The isolation reduces the risk of accidentally publishing the source-built nupkgs themselves rather than the intermediate nupkg, and there are plans to use the isolation for other things later. The isolation implementation (recursive build using `Exec` after appending extra args) is based on https://github.com/dotnet/arcade/issues/5116.

(You can also run `./build.sh` then `./build.sh /p:FromSource=true` without your builds clashing, but this isn't a scenario that's worth truly supporting IMO.)

### DIFF
--- a/eng/Microsoft.DotNet.Arcade.Sdk/tools/SourceBuildArcade.targets
+++ b/eng/Microsoft.DotNet.Arcade.Sdk/tools/SourceBuildArcade.targets
@@ -5,7 +5,6 @@
   -->
 
   <PropertyGroup>
-    <SourceBuildEngineeringDir>$([MSBuild]::NormalizeDirectory('$(RepositoryEngineeringDir)', 'source-build'))</SourceBuildEngineeringDir>
     <SourceBuildArtifactsDir>$([MSBuild]::NormalizeDirectory('$(ArtifactsDir)', 'source-build'))</SourceBuildArtifactsDir>
     <CurrentRepoSourceBuildArtifactsDir>$([MSBuild]::NormalizeDirectory('$(ArtifactsDir)', 'source-build', 'self'))</CurrentRepoSourceBuildArtifactsDir>
     <CurrentRepoSourceBuildArtifactsPackagesDir>$([MSBuild]::NormalizeDirectory('$(CurrentRepoSourceBuildArtifactsDir)', 'packages', '$(Configuration)'))</CurrentRepoSourceBuildArtifactsPackagesDir>
--- a/eng/Microsoft.DotNet.Arcade.Sdk/tools/SourceBuildArcade.targets
+++ b/eng/Microsoft.DotNet.Arcade.Sdk/tools/SourceBuildArcade.targets
@@ -1,0 +1,113 @@
+<Project>
+
+  <!--
+    These targets inject source-build into Arcade's build process.
+  -->
+
+  <PropertyGroup>
+    <SourceBuildEngineeringDir>$([MSBuild]::NormalizeDirectory('$(RepositoryEngineeringDir)', 'source-build'))</SourceBuildEngineeringDir>
+    <SourceBuildArtifactsDir>$([MSBuild]::NormalizeDirectory('$(ArtifactsDir)', 'source-build'))</SourceBuildArtifactsDir>
+    <CurrentRepoSourceBuildArtifactsDir>$([MSBuild]::NormalizeDirectory('$(ArtifactsDir)', 'source-build', 'self'))</CurrentRepoSourceBuildArtifactsDir>
+    <CurrentRepoSourceBuildArtifactsPackagesDir>$([MSBuild]::NormalizeDirectory('$(CurrentRepoSourceBuildArtifactsDir)', 'packages', '$(Configuration)'))</CurrentRepoSourceBuildArtifactsPackagesDir>
+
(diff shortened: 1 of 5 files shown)


eng/Microsoft.DotNet.Arcade.Sdk/tools/SourceBuildArcade.targets
```diff
@@ -0,0 +1,113 @@
+<Project>
+
+  <!--
+    These targets inject source-build into Arcade's build process.
+  -->
+
+  <PropertyGroup>
+    <SourceBuildEngineeringDir>$([MSBuild]::NormalizeDirectory('$(RepositoryEngineeringDir)', 'source-build'))</SourceBuildEngineeringDir>
```
dseefeld, line 8: What is this property used for?  I don't see it anywhere else.

dagood: This is a leftover line from my prototype that I missed deleting--the directory itself doesn't exist either. Good catch, removing.

eng/Microsoft.DotNet.Arcade.Sdk/tools/SourceBuildIntermediate.proj
```diff
@@ -0,0 +1,53 @@
...
+
+    <IsPackable>true</IsPackable>
+    <IsShipping>false</IsShipping>
+    <SuppressDependenciesWhenPacking>true</SuppressDependenciesWhenPacking>
+    <IncludeBuildOutput>false</IncludeBuildOutput>
+
+    <!-- NuGet excludes nupkgs by default: disable this behavior. -->
+    <NoDefaultExcludes>true</NoDefaultExcludes>
+
+    <EnableDefaultSourceBuildIntermediateItems Condition="'$(EnableDefaultSourceBuildIntermediateItems)' == ''">true</EnableDefaultSourceBuildIntermediateItems>
```
dseefeld, line 30: Is the idea here that by default it packages .nupkgs and tar.gzs and if someone wants to override that behavior, they set `EnableDefaultSourceBuildIntermediateItems` to false and then define their own Content item group?

eng/Microsoft.DotNet.Arcade.Sdk/tools/SourceBuildIntermediate.proj
```diff
@@ -0,0 +1,53 @@
...
+
+  <ItemGroup Condition="'$(EnableDefaultSourceBuildIntermediateItems)' == 'true'">
+    <Content Include="$(CurrentRepoSourceBuildArtifactsPackagesDir)**\*.nupkg" PackagePath="artifacts" />
+    <Content Include="$(CurrentRepoSourceBuildArtifactsPackagesDir)**\*.tar.gz" PackagePath="artifacts" />
+  </ItemGroup>
+
+  <Target Name="InitializeSourceBuildIntermediatePackageId"
+          BeforeTargets="GenerateNuspec;InitializeStandardNuspecProperties">
+    <PropertyGroup>
+      <PackageId>Microsoft.SourceBuild.Intermediate.$(GitHubRepositoryName).$(SourceBuildPackageRid)</PackageId>
```
dseefeld, line 44: Is there a version number or commit included in the package name to identify which commit was built?

eng/Microsoft.DotNet.Arcade.Sdk/tools/SourceBuildIntermediate.proj
dagood: Exactly. This is an escape hatch so that repos with duplicated artifacts (e.g. slightly different formats) or unused artifacts can choose to not include them in the intermediate nupkg if it impacts their build.

(The naming is themed to be similar to the default `Compile` item patterns included when building a csproj.)

There's more to work on here (what do we do if the intermediate nupkg is legitimately gigantic) that this PR doesn't include, though.

eng/Microsoft.DotNet.Arcade.Sdk/tools/SourceBuildIntermediate.proj
dagood: `PackageId` is static, there's also a version number that Arcade controls. In a given official build, the version number will generally speaking be shared between the Microsoft packages and source-built intermediate package. For example:

* File: Microsoft.SourceBuild.dotnet-installer.linux-x64.5.0.0-dev.nupkg
  * Nuspec `<id>Microsoft.SourceBuild.dotnet-installer.linux-x64</id>`
  * Nuspec `<version>5.0.0-dev</version>`

PR #130, opened 2020-07-21
Fix Microsoft.AspNetCore.App.Ref/3.1.3 casing: fix DLL production

dagood:
In the targeting pack binlog, I see:

```
MSBuild
    Parameters
        Projects
            /work/src/targetPacks/ILsrc/microsoft.aspnetcore.app.ref/microsoft.aspnetcore.app.ref.proj
            /work/src/targetPacks/ILsrc/Microsoft.AspNetCore.App.Ref/Microsoft.AspNetCore.App.Ref.proj
            ...
        BuildInParallel = True
```

But I only see a project node for the lowercase one. It seems like the MSBuild task de-duplicated the items case-insensitively, causing the one with capitals (3.1.3) to not be built. Our pack uses a batched `Exec` instead, which seems not to de-dup, so the end result is a nupkg that's missing its DLLs.

Fix this by making 3.1.3 consistent with 3.1.2 and friends.

/cc @omajid @directhex


dagood:
This worked in a fresh local build, 3.1.3 filesize is about the same as 3.1.2 and now contains the DLLs.

dseefeld:
Is there something unique to the Fedora build that's causing this?  Because the version built by the last SBRP build doesn't seem to have this problem.  See https://dotnetfeed.blob.core.windows.net/dotnet-core/assets/Private.SourceBuild.ReferencePackages.1.0.0-beta.20367.1.tar.gz   Am I missing something?

dagood:
> Is there something unique to the Fedora build that's causing this?

No, my repro was on CentOS for example.

> See https://dotnetfeed.blob.core.windows.net/dotnet-core/assets/Private.SourceBuild.ReferencePackages.1.0.0-beta.20367.1.tar.gz

Interesting! It is broken, but differently. It has broken 3.0.0, 3.0.1, 3.1.0, and 3.1.2, and a fine 3.1.3. This could be some difference in how the filesystem sorts (or how MSBuild ends up using it), or randomness.

dseefeld:
> Interesting! It is broken, but differently. It has broken 3.0.0, 3.0.1, 3.1.0, and 3.1.2, and a fine 3.1.3. This could be some difference in how the filesystem sorts (or how MSBuild ends up using it), or randomness.

That is interesting!  I only checked 3.1.3 and not the others, because that was the broken one!

PR #139, opened 2020-08-18
Add Arcade SDK yaml template

dseefeld:


azure-pipelines/builds/ci.yml
```diff
@@ -17,68 +17,6 @@ stages:
...
-          condition: succeededOrFailed()
-
-# Based on - https://github.com/dotnet/arcade/blob/9b3f304c7bc9fd4d11be9ca0b466b83e98d2a191/Documentation/CorePackages/Publishing.md#moving-away-from-the-legacy-pushtoblobfeed-task
-- ${{ if and(ne(variables['System.TeamProject'], 'public'), notin(variables['Build.Reason'], 'PullRequest')) }}:
-  - template: /eng/common/templates/post-build/post-build.yml
-    parameters:
-      enableSourceLinkValidation: true
-      #Nuget and Signing Validation are not needed as we only produce a transport package.
-      enableSigningValidation: false
-      enableNugetValidation: false
```
dagood, line 84: Where'd this go?

azure-pipelines/builds/ci.yml
```diff
@@ -17,68 +17,6 @@ stages:
 - stage: build
   displayName: Build
   jobs:
-  - template: /eng/common/templates/jobs/jobs.yml
+  - template: /eng/common/templates/job/build-from-source.yml
```
dagood, line 20: Put this in `/eng/templates/...` to avoid blocking arcade auto-updates? Later we can uptake and delete this copy.

eng/common/templates/job/build-from-source.yml
```diff
@@ -0,0 +1,75 @@
+parameters:
+  jobNamePrefix: 'sourcebuild'    # Override for prefix for job names
+  managedOnly: false              # Specifies whether source is managed only. If so, one build platform is sufficient
+  portableOnly: false             # If true, builds one leg with portable flag enabled
+  prebuiltCheck: true             # Specifies whether to enforce prebuilt checking and fail if new prebuilts are introduced
```
dagood, line 5: Put comments on line before parameter? Matches the existing Arcade SDK template style, more maintainable (alignment!), and IMO more readable.

https://github.com/dotnet/source-build-reference-packages/blob/50dfd943f6e188f47fe138c1c43f2832e1b09555/eng/common/templates/jobs/jobs.yml#L1-L15

eng/common/templates/job/build-from-source.yml
```diff
@@ -0,0 +1,75 @@
...
+  parameters:
+    enablePublishUsingPipelines: true
+    enablePublishBuildArtifacts: true
+    enablePublishBuildAssets: true
+    artifacts:
+      publish:
+        artifacts: true
+        manifests: true
+    jobs:
+      # Only run fedora32 when either managedOnly or portableOnly is true.  Otherwise, run all legs.
```
dagood, line 18: Why?

This also doesn't look correct? (I can change how I'm parsing it to make it correct but it's not how I read it normally...)

eng/common/templates/job/build-from-source.yml
```diff
@@ -0,0 +1,75 @@
...
+        parameters:
+          portableOnly: ${{ parameters.portableOnly }}
+
+    - job: ${{ parameters.jobNamePrefix }}_centos71
+      container: 'mcr.microsoft.com/dotnet-buildtools/prereqs:centos-7-3e800f1-20190501005343'
+      workspace:
+        clean: all
+      steps:
+      - template: ../steps/run-build-from-source.yml
+      condition: and( eq(${{ parameters.portableOnly }}, false), eq( ${{ parameters.managedOnly }}, false))
```
dagood, line 34: Use template condition instead so the ones that don't run don't clutter the AzDO UI?

eng/common/templates/job/build-from-source.yml
```diff
@@ -0,0 +1,75 @@
+parameters:
+  jobNamePrefix: 'sourcebuild'    # Override for prefix for job names
+  managedOnly: false              # Specifies whether source is managed only. If so, one build platform is sufficient
+  portableOnly: false             # If true, builds one leg with portable flag enabled
```
dagood, line 4: Isn't the behavior the inverse? This disables all legs except the one with the portable flag enabled? (And the one with the portable flag should always run, either way, right?)

eng/common/templates/job/build-from-source.yml
```diff
@@ -0,0 +1,75 @@
...
+      - template: ../steps/run-build-from-source.yml
+      condition: and( eq(${{ parameters.portableOnly }}, false), eq( ${{ parameters.managedOnly }}, false))
+
+    - job: ${{ parameters.jobNamePrefix }}_centos8
+      container: 'mcr.microsoft.com/dotnet-buildtools/prereqs:centos-8-daa5116-20200325130212'
+      workspace:
+        clean: all
+      steps:
+      - template: ../steps/run-build-from-source.yml
+      condition: and( eq(${{ parameters.portableOnly }}, false), eq( ${{ parameters.managedOnly }}, false))
```
dagood, line 42: I don't like the duplication here... hard to spot differences. I'd look for something like:

```yml
parameters:
  standardPlatforms:
  - name: 'centos71'
    container: 'mcr.microsoft.com/dotnet-buildtools/prereqs:centos-7-3e800f1-20190501005343'
    # Indicates this is the standard platform: the one that should always build portable.
    standardPlatform: true
  nonPortablePlatforms:
  - name: 'debian9'
    container: 'mcr.microsoft.com/dotnet-buildtools/prereqs:debian-stretch-d61254f-20190807161114'
  - name: 'fedora30'
    container: 'mcr.microsoft.com/dotnet-buildtools/prereqs:fedora-30-38e0f29-20191126135223'

...

- template: ../jobs/build-from-source-jobs.yml
  parameters:
    platforms:
    - ${{ each platform in standardPlatforms }}:
      - ${{ platform }}
    - ${{ if ne(parameters.portableOnly, true) }}:
      - ${{ each platform in nonPortablePlatforms }}:
        - ${{ platform }}
```

eng/common/templates/steps/run-build-from-source.yml
```diff
@@ -0,0 +1,49 @@
+parameters:
+  portableOnly: 'false'
+
+steps:
+  - script: |
```
dagood, line 5: Unnecessary indent level, current style seems to be only indent when needed.

eng/common/templates/steps/run-build-from-source.yml
```diff
@@ -0,0 +1,49 @@
...
+      set -x
+      df -h
+
+      # Setup build arguments
+      _InternalBuildArgs=""
+      if [[ "$BUILD_REASON" != "PullRequest" && "$SYSTEM_TEAMPROJECT" != "public" ]]; then
+        _InternalBuildArgs="/p:DotNetPublishUsingPipelines=true /p:OfficialBuildId=$(BUILD.BUILDNUMBER)"
+      fi
+      if [[ "${{ parameters.portableOnly }}" == "true" ]]; then
+        _InternalBuildArgs="$_InternalBuildArgs /p:ArcadeBuildPortable=true"
```
dagood, line 20: This logic doesn't make sense to me... "only" being necessary for even one job to be portable...

eng/common/templates/steps/run-build-from-source.yml
```diff
@@ -0,0 +1,49 @@
...
+      fi
+      
+      ./build.sh --ci \
+        --configuration $(_BuildConfig) \
+        --restore --build --pack --publish \
+        $_InternalBuildArgs \
+        /p:ArcadeBuildFromSource=true > build.log
```
dagood, line 27: Redirect seems wrong, doesn't this prevent logs from showing up in AzDO?

eng/common/templates/steps/run-build-from-source.yml
```diff
@@ -0,0 +1,49 @@
...
+    displayName: Build
+
+      # Upload build logs for diagnosis.
+  - task: CopyFiles@2
```
dagood, line 31: Indent mismatch, `#` should be on same col as `-`.

eng/common/templates/job/build-from-source.yml
```diff
@@ -0,0 +1,75 @@
+parameters:
+  jobNamePrefix: 'sourcebuild'    # Override for prefix for job names
+  managedOnly: false              # Specifies whether source is managed only. If so, one build platform is sufficient
+  portableOnly: false             # If true, builds one leg with portable flag enabled
+  prebuiltCheck: true             # Specifies whether to enforce prebuilt checking and fail if new prebuilts are introduced
+
+jobs:
+- template: /eng/common/templates/jobs/jobs.yml
```
dagood, line 8: I'm not sure we can use this template like this... it has its own `publish-build-assets`, so if the repo itself uses `jobs.yml`, it might run this twice:

https://github.com/dotnet/source-build-reference-packages/blob/50dfd943f6e188f47fe138c1c43f2832e1b09555/eng/common/templates/jobs/jobs.yml#L46-L48

From a very small sample I see that some repos don't use jobs.yml (runtime, sdk), but others do (xliff-tasks).

I think we need to just use `job.yml`. Repos that use `jobs.yml` can *maybe* plug it in, or we can implement automatic setup in that case and pipe through the right variables. But the general case, I think, is someone using our template as a job(s) template in their build stage, where they've already set up this publish job.

dagood: In SBRP this would look like the `azure-pipelines/builds/ci.yml` still using `jobs.yml`, but also including a source-build job.

And now that I think about it a bit more, I think we do need to make sure our job can be threaded into `jobs.yml` if a repo is using it, because `jobs.yml` sets up the job dependency for `publish-build-assets.yml` using the jobs that are passed into it. (If we don't get source-build in that list, the build may publish before we're done and skip our artifacts.)

azure-pipelines/builds/ci.yml
dseefeld: That will go back at the end of ci.yml.  I just didn't want it to run while I was testing.

azure-pipelines/builds/ci.yml
dseefeld: I can do that.

azure-pipelines/builds/ci.yml
dagood: Don't worry about that--it'll only publish if you have a default channel assignment. (Actually, useful to validate that this is still the case. 😄)

eng/common/templates/job/build-from-source.yml
dseefeld: I would argue that there's not an existing Arcade SDK template style.  They're a bit of a mixed bag.  But I'm fine with changing it.

dagood: Huh, yeah. I looked at a few important-looking once, but just happened to match.

eng/common/templates/job/build-from-source.yml
dseefeld: This isn't something that we need to do, though.  As long as repo owners include our template as a job when including `jobs.yml`, we should be OK, right?  Our template just needs to provide the jobs.

eng/common/templates/steps/run-build-from-source.yml
dseefeld: Fixed

eng/common/templates/steps/run-build-from-source.yml
dseefeld: I removed it.

eng/common/templates/steps/run-build-from-source.yml
dseefeld: Fixed

eng/common/templates/job/build-from-source.yml
dagood: > This isn't something that we need to do, though.

Sorry, what does "this" mean here?

> As long as repo owners include our template as a job when including `jobs.yml`, we should be OK, right?

As is, I think this PR would end up including `jobs.yml` twice. Once when the repo uses it, then again when this source-build template uses it. So they end up with something like this:

* ci.yml
  * jobs.yml
    * [ordinary jobs for Microsoft build]
    * build-from-source.yml
      * jobs.yml
        * centos7 build from source
        * **publish-build-assets.yml** 
    * **publish-build-assets.yml** 

The duplicate is a problem, we should just make a template that provides jobs:

* ci.yml
  * jobs.yml
    * [ordinary jobs for Microsoft build]
    * build-from-source.yml
      * centos7 build from source
    * publish-build-assets.yml 

(I don't actually know if you can use a template like that--to add jobs to the `jobs.yml` template's parameter list. So we might need to modify `jobs.yml` to bridge the gap.)

But we also need to support repos that don't use jobs.yml. They necessarily have `publish-build-assets.yml ` set up already because Arcade requires it, so we shouldn't add it again.

* ci.yml
  * [ordinary jobs for Microsoft build]
  * build-from-source.yml
    * centos7 build from source
  * publish-build-assets.yml

dseefeld: Yes.  That's what I meant when I said, "Our template just needs to provide the jobs."  I just stated it terribly and confusingly! :(

dagood: Hah, ok, perfect. 😄 Without context I thought your comment was meant as justification for the current impl, so I assumed some deep misunderstanding that wasn't there.

eng/common/templates/job/build-from-source.yml
dseefeld: Fixed

eng/common/templates/job/build-from-source.yml
dseefeld: Eliminated duplication

eng/common/templates/steps/run-build-from-source.yml
dseefeld: Corrected this.

eng/templates/job/build-from-source.yml
```diff
@@ -0,0 +1,96 @@
...
+    - job: ${{ parameters.jobNamePrefix }}_${{ platform.name }}
+      ${{ each property in platform.jobProperties }}:
+        ${{ property.key }}: ${{ property.value }}
+      workspace:
+        clean: all
+      steps:
+      - template: ../steps/run-build-from-source.yml
+        parameters:
+          ${{ if eq(platform.include, 'nonPortable') }}:
+            includeNonPortable: true
```
dagood, line 92: Maybe this could pass through all `platform` properties into the parameters via injection to avoid repeating each one?

```yaml
      - template: ../steps/run-build-from-source.yml
        parameters:
          ${{ insert }}: ${{ platform }}
```

(Used this before at https://github.com/dotnet/runtime/blob/master/eng/pipelines/common/platform-matrix-multijob.yml)

To limit what you pass, can do that similar to `jobProperties`, have a `sourceBuildProperties` that's inserted this way.

eng/templates/job/build-from-source.yml
```diff
@@ -0,0 +1,96 @@
...
+  # Specifies that the repo only builds managed components. This indicates that output is not RID-specific 
+  # and only one build job is required. A linux x64 machine will be used, but the build must continue to 
+  # work on all supported platforms equally.
+  managedOnly: false
+  
+  # Specifies that the repo has native components that can link to platform-provided libraries.
+  #   - The job passes /p:ArcadeBuildNonPortable=true to the build to enable non-portable behavior.
+  #   - The job passes a platform-specific TargetRid MSBuild property.
+  #   - Setting this parameter adds additional platform-specific "non-portable" build jobs.
+  includeNonPortable: false
```
dagood, line 14: When I went through the repos, they were all `managedOnly:true` or `managedOnly:false includeNonPortable:true`--I think we can remove `includeNonPortable`. I'm not actually sure when it would ever be used--if someone's building native code, distros prefer to build it non-portable.

eng/templates/job/build-from-source.yml
```diff
@@ -0,0 +1,96 @@
...
+      pool:
+        name: Hosted macOS
+
+  - name: 'ubuntu1804'
+    include: 'nonPortable'
+    targetRID: 'ubuntu.18.04-x64'
+    jobProperties:
+      container: 'mcr.microsoft.com/dotnet-buildtools/prereqs:ubuntu-18.04-f90bc20-20180320154721'
+
+jobs:
```
dagood, line 79: This is a "jobs" template so I think it's in the wrong dir.

eng/templates/steps/run-build-from-source.yml
```diff
@@ -0,0 +1,53 @@
...
+
+  # Specifies the targetRID to pass to the build.
+  targetRID: ''
+
+steps:
+- script: |
+    set -x
+    df -h
+
+    # Setup build arguments
```
dagood, line 13:
```suggestion
    # Set up build arguments
```

eng/templates/steps/run-build-from-source.yml
```diff
@@ -0,0 +1,53 @@
...
+    _InternalBuildArgs=""
+    if [[ "$BUILD_REASON" != "PullRequest" && "$SYSTEM_TEAMPROJECT" != "public" ]]; then
+      _InternalBuildArgs="/p:DotNetPublishUsingPipelines=true /p:OfficialBuildId=$(BUILD.BUILDNUMBER)"
+    fi
+    if [[ "${{ parameters.includeNonPortable }}" == "true" ]]; then
+      _InternalBuildArgs="$_InternalBuildArgs /p:ArcadeBuildNonPortable=true"
+    fi
+    if [[ "${{ parameters.targetRID }}" != "" ]]; then
+      _InternalBuildArgs="$_InternalBuildArgs /p:TargetRid=${{ parameters.targetRID }}"
+    fi
```
dagood, line 24: I think this logic to detect and rephrase the variables can be switched back to variables by using a job template?

eng/templates/job/build-from-source.yml
```diff
@@ -0,0 +1,96 @@
...
+
+- ${{ each platform in parameters.platforms }}:
+  - ${{ if or( eq(platform.include,'always'), and( eq(platform.include, 'notManaged'), eq(parameters.managedOnly, false)), and( eq(platform.include, 'nonPortable'), eq(parameters.includeNonPortable, true))) }}:
```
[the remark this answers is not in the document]
dseefeld: Chopping it up didn't seem to work.  I'm not sure if there's a spacing or character needed to split a line.  But, I implemented your suggestion that `managedOnly` is really the only flag we need, not `includeNonPortable` and this is much more reasonable on one line.

eng/templates/job/build-from-source.yml
dseefeld: Moved.

eng/templates/job/build-from-source.yml
dseefeld: I refactored this and it eliminated the need for this parameter.

eng/templates/job/build-from-source.yml
dseefeld: I refactored, taking this in mind.

eng/templates/steps/run-build-from-source.yml
dseefeld: fixed

eng/templates/steps/run-build-from-source.yml
dseefeld: I switched it back based on combining the steps template and jobs template into one.

dseefeld:
Note: Before merging, I will move the changes to `eng/common/templates/jobs/jobs.yml` to `eng/template/jobs/jobs.yml` to ensure the changes are local to sbrp and won't conflict with arcade updates when these changes are made in arcade.  I thought it was important to see the diff when reviewing the PR.

eng/templates/jobs/build-from-source.yml
```diff
@@ -0,0 +1,130 @@
+parameters:
+  # Specifies the prefix for source-build jobs added to pipeline. Use this if disambiguation is necessary.
+  jobNamePrefix: 'sourcebuild'
+
+  # Specifies that the repo only builds managed components. This indicates that output is not RID-specific 
+  # and only one build job is required. A linux x64 machine will be used, but the build must continue to 
+  # work on all supported platforms equally.
+  # If false, pecifies that the repo has native components that can link to platform-provided libraries.
```
dagood, line 8:
```suggestion
  # If false, specifies that the repo has native components that can link to platform-provided libraries.
```

eng/templates/jobs/build-from-source.yml
```diff
@@ -0,0 +1,130 @@
...
+  #   - Adds additional portable job(s) for supported portable platforms
```
dagood, line 9: I think we should assume the lack of `win` and non-`x64` RIDs is temporary and go with the intent here for readability:

```suggestion
  #   - Adds additional portable jobs for supported portable platforms
```

eng/templates/jobs/build-from-source.yml
```diff
@@ -0,0 +1,130 @@
...
+  #   - Adds additiional non-portable jobs for supported non-portable platforms
```
dagood, line 10:
```suggestion
  #   - Adds additional non-portable jobs for supported non-portable platforms
```

eng/templates/jobs/build-from-source.yml
```diff
@@ -0,0 +1,130 @@
...
+
+  - name: 'ubuntu1804'
+    targetRID: 'ubuntu.18.04-x64'
+    jobProperties:
+      container: 'mcr.microsoft.com/dotnet-buildtools/prereqs:ubuntu-18.04-f90bc20-20180320154721'
+
+jobs:
+
+- ${{ each platform in parameters.platforms }}:
+  - ${{ if or(eq(platform.include,'always'), eq(parameters.managedOnly, false)) }}:
```
dagood, line 74:
```suggestion
  - ${{ if or(eq(platform.include, 'always'), eq(parameters.managedOnly, false)) }}:
```

eng/common/templates/jobs/jobs.yml
```diff
@@ -43,6 +50,12 @@ jobs:
 
       name: ${{ job.job }}
 
+- ${{ if eq(parameters.runBuildFromSource, true) }}:
+  - template: /eng/templates/jobs/build-from-source.yml
```
dagood, line 54: I think we need something in the dependency list to make sure source-build completes before publish to BAR:

https://github.com/dotnet/source-build-reference-packages/blob/89ccfa3d302356ce83125d6d5108662e18f0e2e4/eng/common/templates/jobs/jobs.yml#L61-L70

But I don't know a way to add all source-build jobs in a maintainable way without making this yaml even worse than my earlier suggestions. The list of source-build jobs only exists in the source-build template, so we'd need to rearrange stuff and make some extra templates to get it across... I can't stand that when we want to do the tiniest bit of data reuse, AzDO puts us on this convoluted path.

---

I think we can escape.... we can include a NOP/sentinel build job in our source-build template that has a consistent name and depends on every source-build job. Then in jobs.yml, we'd add a dependency on that sentinel job if source-build is enabled. Repos that don't use jobs.yml can simply use that sentinel job's consistent name. Waste of time, since AFAIK AzDO doesn't has server jobs yet that would let us avoid spinning up an actual machine to "do" the NOP. But at the moment it seems like the best we can do for reusability/maintainability.

To make it less wasteful for managed-only repos, we could use the sentinel job in a flexible way rather than *always* spinning up a specifically-named job for every target:
* Managed-only:
  * `source-build-main` (runs "centos71_portable" but uses the sentinel job name, so others can still add dependencies on it)
* Non-manged-only:
  * `source-build-centos71_portable`
  * `source-build-macOS_portable`
  * `source-build-centos71`
  *  ...,
  * `source-build-main` (does nothing, has dependencies on all previous jobs)

dagood: > AFAIK AzDO doesn't has server jobs yet

Oh: https://docs.microsoft.com/en-us/azure/devops/pipelines/process/phases?view=azure-devops&tabs=yaml#server-jobs

Can probably keep it simple and always have one of those as the sentinel.

dseefeld: Thanks for catching that.  I updated with a sentinel job for source-build and setup the dependency.  I tested it in my test repo and it seems to work great.  I also updated job/display names to better match the arcade standard for their AzDo jobs.

eng/templates/jobs/build-from-source.yml
```diff
@@ -0,0 +1,138 @@
...
+        name: Hosted macOS
+
+  - name: 'Ubuntu1804'
+    targetRID: 'ubuntu.18.04-x64'
+    jobProperties:
+      container: 'mcr.microsoft.com/dotnet-buildtools/prereqs:ubuntu-18.04-f90bc20-20180320154721'
+
+jobs:
+- job: Build_From_Source_Main
+  displayName: Build From Source Main
```
dagood, line 73: I see I suggested "main", but not sure why... now that I see it here, I think we should go with `Complete` to make that string more obvious when it shows up in a list of job dependencies.

On the display name side, I'm also wondering if "Source-build (x)" would be more concise and if we should stick with that name in "branding" on the source-build effort. Seems more familiar, e.g. the effort's called arcade-powered source-build.

image

image

eng/templates/jobs/build-from-source.yml
```diff
@@ -0,0 +1,138 @@
...
+      workspace:
+        clean: all
+
+      variables:
+      - name: _BuildConfig
+        value: Release
+      - name: _InternalBuildArgs
+        value: ''
+      - name: _NonPortableBuildArgs
+        value: ''
```
dagood, line 95: Nit: trailing space

eng/templates/jobs/build-from-source.yml
```diff
@@ -0,0 +1,138 @@
...
+      container: 'mcr.microsoft.com/dotnet-buildtools/prereqs:ubuntu-18.04-f90bc20-20180320154721'
+
+jobs:
+- job: Build_From_Source_Main
+  displayName: Build From Source Main
+  pool: server
+  dependsOn:
+  - ${{ each platform in parameters.platforms }}:
+    - ${{ if or(eq(platform.include, 'always'), eq(parameters.managedOnly, false)) }}:
+      - ${{ parameters.jobNamePrefix }}_${{ platform.name }}
```
dagood, line 78: Nit: trailing space

eng/templates/jobs/build-from-source.yml
```diff
@@ -0,0 +1,138 @@
...
+        - name: _NonPortableBuildArgs
+          value: >-
+            /p:ArcadeBuildNonPortable=true
+            /p:TargetRid=${{ platform.targetRID }}
+
+      steps:
+      - script: |
+          set -x
+          df -h
+          ./build.sh --ci \
```
dagood, line 111: Something I realized we'll definitely need is a way to change the root script, for repos that aren't onboarded to an Arcade entrypoint like NuGet.Client. (Maybe this is only NuGet.Client? Not sure.) Then they can change it to an `eng/source-build.sh` or whatever, and we don't have to worry about letting people also configure the args in this template.

(Defaulting to `./build.sh` seems good though.)

eng/templates/jobs/build-from-source.yml
dseefeld: That makes sense.  What do you think about the name of the script itself?  `build-from-source.yml` -> `source-build.yml`?

eng/templates/jobs/build-from-source.yml
dseefeld: Thanks.  I added a parameter called `buildScript` and initialized it to `./build.sh`.

eng/templates/jobs/jobs.yml
```diff
@@ -0,0 +1,87 @@
...
+
+- ${{ if eq(parameters.runSourceBuild, true) }}:
+  - template: /eng/templates/jobs/source-build.yml
+    parameters:
+      ${{ each parameter in parameters.sourceBuildParameters }}:
+        ${{ parameter.key }}: ${{ parameter.value }}
+
+- ${{ if and(eq(parameters.runAsPublic, 'false'), ne(variables['System.TeamProject'], 'public'), notin(variables['Build.Reason'], 'PullRequest')) }}:
+  - ${{ if or(eq(parameters.enablePublishBuildAssets, true), eq(parameters.artifacts.publish.manifests, 'true'), ne(parameters.artifacts.publish.manifests, '')) }}:
+    - template: ../job/publish-build-assets.yml
```
dagood, line 61: Looks like this should be an absolute path for now, official build failure:
https://dev.azure.com/dnceng/internal/_build/results?buildId=790988&view=results

PR #144, opened 2020-09-02
Add packages to support Preview8 build

dseefeld:
Add the following packages:

  - System.ValueTuple,4.5.0
  - System.Text.Encoding.CodePages,4.5.1
  - Microsoft.Build,16.6.0
  - Microsoft.Build.Utilities.Core,16.6.0
  - Microsoft.Build.Framework,16.6.0  

These also required the following dependent projects:
  - System.Security.Principal.Windows,4.7.0
  - System.Security.AccessControl,4.7.0
  - System.Security.Permissions,4.7.0


src/referencePackages/src/microsoft.build.utilities.core/16.6.0/microsoft.build.utilities.core.nuspec
```diff
@@ -0,0 +1,36 @@
...
+<package xmlns="http://schemas.microsoft.com/packaging/2013/05/nuspec.xsd">
+  <metadata>
+    <id>Microsoft.Build.Utilities.Core</id>
+    <version>16.6.0</version>
+    <authors>Microsoft</authors>
+    <owners>Microsoft</owners>
+    <requireLicenseAcceptance>true</requireLicenseAcceptance>
+    <license type="expression">MIT</license>
+    <licenseUrl>https://licenses.nuget.org/MIT</licenseUrl>
+    <icon>MSBuild-NuGet-Icon.png</icon>
```
dagood, line 11: Share the pngs like the keys rather than duplicating them? (Also share with targeting packs, which already have a few duplicates?)

dagood (approved):

crummel (approved):

src/referencePackages/src/microsoft.build.utilities.core/16.6.0/microsoft.build.utilities.core.nuspec
dseefeld: I'm not sure it makes sense to do that in the pre-arcade branch because it will require additional infra to copy icon files to the output dirs that we use to pack.  It doesn't seem worth it for a handful of png files.  I think this would make sense in the master branch since the build is updated there.

dagood: That's fair. :+1:

PR #152, opened 2020-09-11
Fix text-only nuspec reference to override default; remove unused pkgproj

dagood:
Set the `NuspecFile` property when `PackNuspecInPlace=true`, not just `false`. This fixes https://github.com/dotnet/source-build/issues/1742, among other problems: the nuspecs were the Arcade-generated defaults so the packages were different in other ways, too.

When I searched for `NuspecFile` in the project, I found a `.pkgproj` file that isn't used--removed it too.


src/Directory.Build.targets
```diff
@@ -103,18 +111,22 @@
...
-
     <PropertyGroup>
       <NuspecFile>$(TFMPackTempOutputDir)@(NuspecFile)</NuspecFile>
     </PropertyGroup>
   </Target>
 
+  <Target Name="SetManualPackTaskInputsInPlace"
+          Condition="'$(PackNuspecInPlace)' == 'true'"
+          DependsOnTargets="FindNuspecFileItem"
+          BeforeTargets="_CalculateInputsOutputsForPack;GenerateNuspec">
```
dagood, line 124: Wait, this is silly, I should just move the condition to the propertygroup.

dseefeld (approved):

crummel (approved):

PR #160, opened 2020-09-18
Support RID-specific arcade-powered source-build with jobs template

dagood:
Use https://github.com/dotnet/arcade/pull/6197 to do RID-specific builds, or managed only builds that have no suffix. This is in support of moving the template to dotnet/arcade: https://github.com/dotnet/source-build/issues/1699.

This repo is managed-only, so it looks like https://dev.azure.com/dnceng/internal/_build/results?buildId=820925&view=results:

> image

A RID-specific build must provide its own platform configuration. It can't be passed down by Arcade, because then we have no way to gradually roll out RID adds/removes without also blocking Arcade SDK updates. For example, this:

```yaml
      runSourceBuild: true
      sourceBuildParameters:
        platforms:
        - name: 'Centos71_Portable'
          container: 'mcr.microsoft.com/dotnet-buildtools/prereqs:centos-7-3e800f1-20190501005343'
        - name: 'MacOS_Portable'
          pool: { vmImage: 'macOS-10.14' }
        - name: 'Centos71'
          nonPortable: true
          container: 'mcr.microsoft.com/dotnet-buildtools/prereqs:centos-7-3e800f1-20190501005343'
        - name: 'Centos8'
          nonPortable: true
          container: 'mcr.microsoft.com/dotnet-buildtools/prereqs:centos-8-daa5116-20200325130212'
        - name: 'Debian9'
          nonPortable: true
          container: 'mcr.microsoft.com/dotnet-buildtools/prereqs:debian-stretch-20200918130533-047508b'
        - name: 'Fedora30'
          nonPortable: true
          container: 'mcr.microsoft.com/dotnet-buildtools/prereqs:fedora-30-38e0f29-20191126135223'
        - name: 'Fedora32'
          nonPortable: true
          container: 'mcr.microsoft.com/dotnet-buildtools/prereqs:fedora-32-20200512010618-163ed2a'
        - name: 'MacOS'
          nonPortable: true
          pool: { vmImage: 'macOS-10.14' }
        - name: 'Ubuntu1804'
          nonPortable: true
          container: 'mcr.microsoft.com/dotnet-buildtools/prereqs:ubuntu-18.04-20200918145614-047508b'
```

Produces https://dev.azure.com/dnceng/internal/_build/results?buildId=820919&view=results:

> image

(That's just for testing, this repo should never produce RID-specific packages. In fact, I think very few repos will, maybe only dotnet/runtime, dotnet/aspnetcore, and dotnet/installer.)

### DIFF
--- a/eng/templates/jobs/source-build.yml
+++ b/eng/templates/jobs/source-build.yml
@@ -6,23 +6,27 @@ parameters:
   # necessary. This also changes the ID of 'Source_Build_Complete'.
   jobNamePrefix: 'Source_Build'
 
-  # A container to use. Runs in docker.
-  container: ''
-
-  # A pool to use. Runs directly on an agent.
-  pool: null
-
-  # A list of job properties to inject at the top level. Only 'container' and 'pool' are expected so
-  # far (so we provided shortcut parameters for those in particular), but this is left in for
-  # potential extensibility.
-  jobProperties: {}
-
   # Specifies the build script to invoke to perform the build in the repo. The default should work
   # for typical Arcade repositories, but this is customizable for difficult situations.
   buildScript: './build.sh'
 
-  # Defines the platforms on which to run each build job. By default, runs one job on a linux-x64
-  # machine, suitable for managed-only repositories.
+  # Defines the platforms on which to run build jobs. By default, one job on a linux-x64 machine,
+  # suitable for managed-only repositories. This is an array of objects with these properties:
+  #
+  # name: ''
+  #   The name of the job, included in the ID.
+  # targetRID: ''
+  #   The name of the target RID to use, instead of the one auto-detected by Arcade.
+  # nonPortable: false
+  #   Enables non-portable mode. This means a more specific RID (e.g. fedora.32-x64 rather than
+  #   linux-x64), and compiling against distro-provided packages rather than portable ones.
+  # container: ''
+  #   A container to use. Runs in docker.
+  # pool: {}
+  #   A pool to use. Runs directly on an agent.
+  # jobProperties: {}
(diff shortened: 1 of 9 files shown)


eng/templates/jobs/source-build.yml
```diff
@@ -1,139 +1,103 @@
...
+          --configuration $(_BuildConfig) \
+          --restore --build --pack --publish \
+          $(_InternalBuildArgs) \
+          $(_TargetRidArgs) \
+          /p:SourceBuildNonPortable=${{ platform.nonPortable }} \
+          /p:ArcadeBuildFromSource=true
+      displayName: Build
+
+    # Upload build logs for diagnosis.
+    - task: CopyFiles@2
```
dseefeld, line 84: Not necessarily specific to this PR, but do you think we want to upload the intermediate nupkg for non-official builds?  It seems like it might be helpful for debugging issues.

dseefeld (approved):

crummel (approved):

eng/templates/jobs/source-build.yml
dagood: It would, but the repos don't even do that with their own artifacts today, so I expect they won't be ok adding this for source-build. (I've seen some that didn't even save binlogs, which is somewhat terrifying and I won't go that far.) I haven't heard of any particular storage constraints but it would be a *lot* of data (GB per build, depending on repo--on every single PR).

With arcade-powered source-build I'm hoping that it's quickly reproducible enough that people can follow similar approaches for debugging bad artifacts as they do with the existing lack of package artifacts saved in PR validation.

PR #162, opened 2020-09-22
Use common templates from Arcade

dagood:
Delete the temp templates and use the source-build ones from Arcade.


dagood:
Going ahead with merge, this is a trivial update + switchover.

PR #164, opened 2020-09-24
Add System.IO.PipesAccessControl,4.5.1 and System.Resources.Extensions,4.6.0

dseefeld:


dagood (approved):

dseefeld:
I also reverted the changes to add System.Text.Json and related packages because these have not been tested in a tarball, yet.  Microsoft.Bcl.AsyncInterfaces is currently causing problems with my tarball build locally.  By removing these, it will show up as a prebuilt, but I'll re-address that with prebuilt removal later after I get the tarball building through with these other reference packages.

PR #166, opened 2020-10-01
Add new MS.NETCore.Platforms and Targets.

crummel:


dseefeld:
> > ```
> > .nupkg.metadata
> > .signature.p7s
> > ```
> 
> Should we be including these? Seems like they could cause a checksum mismatch or signing error.

Agreed. These should be removed.  They're not include with other textOnlyPackages, either.

dagood (approved):

PR #167, opened 2020-10-01
Add NUnit template packages to text-only packages

crummel:


dagood:
Hmm. I'm not a fan of adding vs. the alternatives. Note that it's not a .NET Foundation/Microsoft package. (For future reference, these templates came into the product via https://github.com/nunit/dotnet-new-nunit/issues/13.)

Alternatives:

* Add https://github.com/nunit/dotnet-new-nunit?
* Don't bundle it, and allow users to use `dotnet new -i NUnit3.DotNetNew.Template` if they want it as directed on https://github.com/nunit/dotnet-new-nunit?

I know I've poked a lot of times about why we got rid of the text-only package decomposition-recomposition, but this is an example of where that automation shines. It lets us keep redisting this text-only stuff without worrying about maintaining it so intentionally and bloating this repo even further.

@dseefeld

dseefeld:
What we got rid of was the reference-only package IL decomposition/recomposition.  Text-only packages was a bit of a side effect of that.  But, we could add back just the text-only package part of it.  My main concern with that is that our long term goal is to be able to generate a tarball without doing a production build and that would move away from that goal.  I would rather have us add the repo.

dagood:
> But, we could add back just the text-only package part of it. My main concern with that is that our long term goal is to be able to generate a tarball without doing a production build and that would move away from that goal.

Something I think is interesting here is that arcade-powered source-build is caching the production build: we could do text-only prebuilt decomposition during the official build and include the text-only layout in the intermediate nupkgs. If we still have to download some previously-source-built bits to make the tarball buildable, downloading intermediate nupkgs isn't much different.

But actually beyond that, we *could* do text-only prebuilt decomposition (even ref-only decomposition) without a production build at all. https://github.com/dotnet/source-build/issues/831 lays out a way to do that.

On the other hand, the less infra, the better, and if we can just keep adding repos, that sounds best.

eng/Build.props
```diff
@@ -4,7 +4,7 @@
     <TargetingPackageProject Include="$(RepoRoot)src\targetPacks\ILsrc\**\*.csproj" />
     <ProjectToBuild Include="@(TargetingPackageProject)" />
 
-    <TextOnlyPackageProject Include="$(RepoRoot)src\textOnlyPackages\src\**\*.csproj" />
+    <TextOnlyPackageProject Include="$(RepoRoot)src\textOnlyPackages\src\**\*.csproj" Exclude="$(RepoRoot)src\textOnlyPackages\src\**\Content\**\*.csproj" />
```
dagood, line 7: I think clamping down the `Include` is little more maintainable/readable than adding wildcard `Exclude`s:

```suggestion
    <TextOnlyPackageProject Include="$(RepoRoot)src\textOnlyPackages\src\*\*\*.csproj" />
```

dagood (approved):
I think it might be fine to accept this for now, I realized https://github.com/nunit/dotnet-new-nunit doesn't have a lot of changes over time so I wouldn't expect this particular text-only package to be a problem.

eng/Build.props
crummel: Yeah, that does seem better, thanks.

crummel:
I logged https://github.com/dotnet/source-build/issues/1786 to look into adding the repo and noted that we should remove these if we do.  Sound okay for now?

dagood:
Sounds good to me, thanks for filing. 👍

PR #170, opened 2020-10-20
Add a set of prebuilts that are reported as DirectDependencies

dseefeld:
Microsoft.Win32.Registry,4.7.0
System.Diagnostics.DiagnosticSource,4.7.0
System.Data.Common,4.3.0
System.Net.NameResolution,4.3.0
System.Reflection.TypeExtensions,4.7.0
System.Text.Encoding.CodePages,4.7.0
System.Data.SqlClient,4.8.1
System.Diagnostics.Contracts,4.3.0
System.IO.UnmanagedMemoryStream,4.3.0
System.IO.MemoryMappedFiles,4.3.0
System.Runtime,4.3.1
System.Collections.Immutable,1.6.0
System.Collections.Immutable,1.7.0
System.Reflection.Metadata,1.7.0
System.Reflection.Metadata,1.8.0


src/referencePackages/src/microsoft.win32.primitives/4.3.0/Microsoft.Win32.Primitives.csproj
```diff
@@ -15,7 +15,6 @@
     <Compile Include="**/ref/$(TargetFramework)/*.cs" />
     <Compile Include="**/lib/$(TargetFramework)/*.cs" />
   </ItemGroup>
-
     <ItemGroup Condition=" '$(TargetFramework)' == 'netstandard1.3' ">
```
dagood, line 18: Stray whitespace change? This blank section separator is one that I actually expect to be there.

(On the other hand, it would be nice to do some cleanup of the mixed 2-space vs. 4-space tabs and stuff like that, but without tooling being tied to the repo version (so we can fix the source of the formatting issues as well) it's probably not the right time to try that.)

src/referencePackages/src/system.reflection.primitives/4.3.0/System.Reflection.Primitives.csproj
```diff
@@ -15,12 +15,10 @@
     <Compile Include="**/ref/$(TargetFramework)/*.cs" />
     <Compile Include="**/lib/$(TargetFramework)/*.cs" />
   </ItemGroup>
-
     <ItemGroup Condition=" '$(TargetFramework)' == 'netstandard1.0' ">
```
dagood, line 18: More whitespace changes. (Skimming through--there may be more I didn't mark.)

dagood (approved):

src/referencePackages/src/microsoft.win32.primitives/4.3.0/Microsoft.Win32.Primitives.csproj
dseefeld: I updated the generator to generate the projects and corrected the 2-space vs. 4-space issue as well as spacing between sections.  Now all generated projects will have 2-space tabs with one space between sections.  This will only be for projects that are new or are re-generated because they are dependencies.

PR #176, opened 2020-11-05
Add System.Data.SqlClient,4.8.2

dseefeld:


dagood:
Why add this rather than unify on 4.8.1? https://github.com/dotnet/source-build/pull/1858 I already did that since I assigned myself on the table. 😕

PR #177, opened 2020-11-09
Add more web project templates.

crummel:


crummel:
This can be mass-ported later.

PR #180, opened 2020-11-11
Updated SBRP repo to use 5.0 SDK & 5.0 Arcade

dseefeld:
Also adds text-only packages for Microsoft.NETFramework.ReferenceAssemblies,1.0.0 & 1.0.0-preview.2

### DIFF
--- a/eng/Version.Details.xml
+++ b/eng/Version.Details.xml
@@ -3,7 +3,7 @@
   <ProductDependencies>
   </ProductDependencies>
   <ToolsetDependencies>
-    <Dependency Name="Microsoft.DotNet.Arcade.Sdk" Version="1.0.0-beta.20113.5">
+    <Dependency Name="Microsoft.DotNet.Arcade.Sdk" Version="5.0.0-beta.20431.1">
       <Uri>https://github.com/dotnet/arcade</Uri>
       <Sha>15f00efd583eab4372b2e9ca25bd80ace5b119ad</Sha>
     </Dependency>
--- a/global.json
+++ b/global.json
@@ -1,6 +1,6 @@
 {
   "tools": {
-    "dotnet": "5.0.100-rc.1.20452.10"
+    "dotnet": "5.0.100"
   },
   "msbuild-sdks": {
     "Microsoft.DotNet.Arcade.Sdk": "1.0.0-beta.20113.5"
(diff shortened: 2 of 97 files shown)


dagood:
Note to future viewers: the CI trigger failed on this PR--the web hook payload was sent and GitHub got a 200 response back, but no build started. To work around this, we pushed the branch to the dotnet/source-build-references-packages repo and triggered the build manually. The results apply to the commit, which is why it's showing up now--the automatic trigger didn't ever work.

dagood:
1. Why rc1, not GA?

```diff
--- a/global.json
+++ b/global.json
@@ -1,6 +1,6 @@
 {
   "tools": {
-    "dotnet": "3.1.106"
+    "dotnet": "5.0.100-rc.1.20452.10"
   },
```

---

2. Why are these necessary? Don't we have a `sed` expression that should be doing this? We should very strongly favor autoamted replacement rather than patches IMO.

```diff
   .method /*060000F8*/ public hidebysig specialname rtspecialname 
-          instance void  .ctor(string constraint) cil managed
+          instance void  .ctor(string 'constraint') cil managed
   // SIG: 20 01 01 0E
   {
```

dseefeld:
> We should very strongly favor automated replacement rather than patches IMO.
I'm not sure why we want to include a sed command to make this change rather than changing the code once?  This isn't a patch that we need to maintain.

dseefeld:
> Why rc1, not GA?

I'll update to GA.

dagood:
> This isn't a patch that we need to maintain.

If we ever need to add another version of these packages, or any additional packages that require the same change, we need to apply the same fix. If we have tooling we can use to apply that fix (the `sed`), we should simply put that tooling in the build so the manual process is as small as possible.

dagood:
I'm calling it patches because this is a set of changes applied on top of the actual "truth" text output of `ildasm`. If we were to get new `ildasm` output, we would need to reapply the patches. I think it's an accurate analogy at least.

(There's potentially an argument about *patches* vs. *patch files* but I think *patches* is probably just an ambiguous term here. 😄)

dseefeld:
>>This isn't a patch that we need to maintain.

>If we ever need to add another version of these packages, or any additional packages that require the same change, we need to apply the same fix. If we have tooling we can use to apply that fix (the sed), we should simply put that tooling in the build so the manual process is as small as possible.

Good enough.  I'll remove them.

dagood (approved):

PR #182, opened 2020-11-11
Nothing.

dagood:
No files changed. Let's see if AzDO CI triggers.


dagood:
I guess 0 works but ∞ (https://github.com/dotnet/source-build-reference-packages/pull/180) doesn't. 🤔 Thought it might be the web hook payload `"changed_files": 0` since it uses 0 for both cases, but I guess not.

PR #183, opened 2020-11-11
Add set of referenceAssemblies (net40,net45,net46,net461) v1.0.0

dseefeld:


dagood (approved):
Poked through it in my Git client, seems good.

crummel:
image
Wow, can't say I've seen that before.  Also looking through this locally.

PR #184, opened 2020-11-11
Add set of referenceAssemblies (net462,net47,net472,net48) v1.0.0

dseefeld:


dagood (approved):
Poked through it in my Git client, seems good.

PR #189, opened 2020-11-16
Revert "Add Microsoft.Win32.System.Events,4.7.0 & System.Drawing.Common,4.7.0 (#187)"

dseefeld:
This reverts commit 12c32bc4f0c23b77db106e5659c42ff471209993.

System.Drawing.Common,4.7.0 is causing an issue with the tarball build:

```
/tb/tarball_886198/src/installer.5044b938294b7b920a5a4393a5f0f8b7b1750548/src/redist/targets/Crossgen.targets(127,5): error : Error: Could not load file or assembly 'System.Drawing.Common, Version=4.0.2.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51'. The system cannot find the file specified. [/tb/tarball_886198/src/installer.5044b938294b7b920a5a4393a5f0f8b7b1750548/src/redist/redist.csproj] 
/tb/tarball_886198/src/installer.5044b938294b7b920a5a4393a5f0f8b7b1750548/src/redist/targets/Crossgen.targets(127,5): error : (0x80070002) [/tb/tarball_886198/src/installer.5044b938294b7b920a5a4393a5f0f8b7b1750548/src/redist/redist.csproj] 
```

Reverting this package until a solution can be determined.


dagood (approved):

dagood:
Looking at this because it ends up overlapping with `Microsoft.Win32.SystemEvents/4.7.0`. (Anything built in dotnet/runtime with a recent version that is at some point used as a lib/runtime, probably.)

I believe that if we have MSBuild/Roslyn/FSharp upgrade their `System` packages, we avoid missing the lib/runtime DLL. But then we have to upgrade them to target `net5.0` (from ~`netcoreapp3.1`) because we only build `net5.0` in dotnet/runtime. This might let us avoid this PR entirely.

An alternative is to have dotnet/sdk override the `System` packages with a top-level dependency on the 5.0.0 source-built version. At that point we're already targeting `net5.0`, so it should take the real lib/runtime fine. This way we can get these ref-only packages back into SBRP but not break the build.

PR #191, opened 2020-11-19
Add System.Drawing.Common,4.7.0 & Microsoft.Win32.SystemEvents,4.7.0

dseefeld:
Microsoft.NETCore.Platforms,3.1.0 is a required dependency


dagood (approved):
> Microsoft.NETCore.Platforms,3.1.0 is a required dependency

This was a little cryptic to me, so to spell it out for future reference: this PR also includes the `Microsoft.NETCore.Platforms/3.1.0` text-only package.

PR #192, opened 2020-11-20
Add Docker and IL text-only SDKs

dagood:
Microsoft.Docker.Sdk/1.1.0
Microsoft.NET.Sdk.IL/5.0.0-preview.8.20359.4


src/textOnlyPackages/src/microsoft.docker.sdk/1.1.0/Microsoft.Docker.Sdk.csproj
```diff
@@ -0,0 +1,12 @@
+<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <ProjectNameToLower>$(MSBuildProjectName.ToLowerInvariant())</ProjectNameToLower>
+    <Version>$([System.IO.Path]::GetFileName('$(MSBuildProjectDirectory)'))</Version>
```
dagood, line 4: Note, removed the duplicate hard-coded name + version, so this `csproj` can be copied freely for new text-only prebuilts and only the filename needs to be adjusted.

(Could go a step further and make this work like `master`, but avoiding broader changes for now.)

/cc @dseefeld

dseefeld (approved):

PR #193, opened 2020-11-22
Add 4.7.0 library extensions packages

dagood:
```
System.ComponentModel.Annotations,4.7.0
System.Security.Cryptography.Cng,4.7.0
System.Security.Cryptography.Pkcs,4.7.0
System.Security.Cryptography.Xml,4.7.0
System.Windows.Extensions,4.7.0,netcoreapp3.0
```


dagood:
With this fix:

```xml
  <!--
    If netstandard1.6 supports TargetFramework (that is, if TargetFramework <= netstandard1.6),
    disable the implicit NETStandard.Library reference. If we allow it, restore fails due to a
    circular reference. The cycle is because this package is part of .NET Standard.
  -->
  <PropertyGroup Condition="$([MSBuild]::IsTargetFrameworkCompatible('netstandard1.6', '$(TargetFramework)'))">
    <DisableImplicitFrameworkReferences>true</DisableImplicitFrameworkReferences>
  </PropertyGroup>
```

I noticed that the sbrp gen tooling seems to be trying to figure out whether it should add `<DisableImplicitFrameworkReferences>true</DisableImplicitFrameworkReferences>` globally depending on what TFMs are present, but maybe this is better to solve on the SBRP side, like this.

crummel (approved):

dseefeld (approved):

PR #197, opened 2020-11-24
Update PrivateSourceBuiltArtifacts to a 5.0 version

dseefeld:
Update PrivateSourceBuiltArtifacts to a recent 5.0 version.  Update IL src to include the same fixup that is present in the master branch to make 'constraint' a keyword.  This allows it to be built with 5.0 ilasm.  

Also, since NETStandard.Library is no longer built in 5.0, the package version is not available in PackageVersions.props, so it needs to be included in dir.props for reference packages.

### DIFF
--- a/src/referencePackages/src/dir.props
+++ b/src/referencePackages/src/dir.props
@@ -30,6 +30,6 @@
       NETStandard.Library is no longer built in 5.0, so the version does not exist in
       PackageVersions.props.  Setting to latest version built in sbrp.
     -->
-    <NETStandardLibraryPackageVersion>2.2.0-prerelease.19564.1</NETStandardLibraryPackageVersion>
+    <NETStandardLibraryPackageVersion>2.0.3</NETStandardLibraryPackageVersion>
   </PropertyGroup>
 </Project>
--- a/src/referencePackages/src/dir.props
+++ b/src/referencePackages/src/dir.props
@@ -26,5 +26,10 @@
     <EnableDefaultCompileItems>false</EnableDefaultCompileItems>
     <GenerateAssemblyInfo>false</GenerateAssemblyInfo>
     <WarningsAsErrors></WarningsAsErrors>
+    <!--
+      NETStandard.Library is no longer built in 5.0, so the version does not exist in
+      PackageVersions.props.  Setting to latest version built in sbrp.
+    -->
+    <NETStandardLibraryPackageVersion>2.2.0-prerelease.19564.1</NETStandardLibraryPackageVersion>
   </PropertyGroup>
 </Project>
--- a/src/targetPacks/assemble.targets
+++ b/src/targetPacks/assemble.targets
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
-<Project DefaultTargets="AssembleTargetPack" ToolsVersion="15.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+<Project ToolsVersion="15.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <ItemGroup>
       <!--
         Files that may need the 'constraint' token fixed before they can be assembled. They were
@@ -74,7 +74,7 @@
   </Target>
 
   <Target Name="AssembleTargetPack"
-          Condition="'@(TargetingPackSrcNotRequiringConstraintPatch)' != '' and '$(SkipTargetingPacks)' != 'true' ">
+          Condition="'@(TargetingPackSrc)' != '' and '$(SkipTargetingPacks)' != 'true' ">
 
     <PropertyGroup>
       <BuildInParallel Condition="'$(BuildInParallel)' == ''">true</BuildInParallel>
--- a/src/targetPacks/assemble.targets
+++ b/src/targetPacks/assemble.targets
@@ -1,5 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="15.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+
   <ItemGroup>
       <!--
         Files that may need the 'constraint' token fixed before they can be assembled. They were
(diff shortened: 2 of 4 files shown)


src/referencePackages/src/dir.props
```diff
@@ -26,5 +26,10 @@
     <EnableDefaultCompileItems>false</EnableDefaultCompileItems>
     <GenerateAssemblyInfo>false</GenerateAssemblyInfo>
     <WarningsAsErrors></WarningsAsErrors>
+    <!--
+      NETStandard.Library is no longer built in 5.0, so the version does not exist in
+      PackageVersions.props.  Setting to latest version built in sbrp.
+    -->
+    <NETStandardLibraryPackageVersion>2.2.0-prerelease.19564.1</NETStandardLibraryPackageVersion>
```
dagood, line 33: Why does this version make more sense than the latest that was actually shipped to the gallery, 2.0.3? (What is this used for?)

src/targetPacks/assemble.targets
```diff
@@ -1,19 +1,90 @@
...
+      <TargetingPackSrc
+        Include="$(MSBuildProjectDirectory)\**\*.il"
+        Exclude="@(TargetingPackSrcRequiringConstraintPatch)" />
   </ItemGroup>
-  <Target Name="AssembleTargetPack" Condition="'@(TargetingPackSrc)' != '' and '$(SkipTargetingPacks)' != 'true' ">
+
+  <Target Name="GetProjectSrc"
+          BeforeTargets="PatchIl">
+    <PropertyGroup>
+      <PatchedILSrcTemp>$(RepoRoot)artifacts/patchedILSrc/</PatchedILSrcTemp>
```
dagood, line 41: IMO this prop name should end in `Dir` for convention.

src/targetPacks/assemble.targets
```diff
@@ -1,19 +1,90 @@
...
+    </PropertyGroup>
+    <ItemGroup>
```
dagood, line 43: Missing newline for readability.

src/targetPacks/assemble.targets
```diff
@@ -1,19 +1,90 @@
...
+      <TargetingPackSrc
+        ILSrcFile="%(FullPath)"
+        RelativeOutputAssemblyFile="%(RecursiveDir)%(Filename).dll" />
+      <TargetingPackSrcRequiringConstraintPatch
+        ILSrcFile="$(PatchedILSrcTemp)$(MSBuildProjectName)/%(RecursiveDir)%(Filename).patched%(Extension)"
+        RelativeOutputAssemblyFile="%(RecursiveDir)%(Filename).dll"
+        DestinationFile="$(PatchedILSrcTemp)$(MSBuildProjectName)/%(RecursiveDir)%(Filename).patched%(Extension)" />
+      <AllSrc
```
dagood, line 51: I'd suggest pulling descriptiveness up if the separate item lists are really necessary (rather than just putting the items into `TargetingPackSrc` like `master`):

`TargetingPackSrc`, `TargetingPackSrcRequiringConstraintPatch`, `AllSrc` ->
`TargetingPackSrcNotRequiringConstraintPatch`, `TargetingPackSrcRequiringConstraintPatch`, `TargetingPackSrc`

src/targetPacks/assemble.targets
```diff
@@ -1,19 +1,90 @@
...
+    <!--
+      Wrap constraint in single quotes in certain circumstances to avoid using reserved token.
+      Function param:
+        " constraint," => " 'constraint',"
+        " constraint)" => " 'constraint')"
+        " constraint"(eol) => " 'constraint'"(eol) (%24 is $ escaped in hex)
+    -->
+    <Exec Command="
+      set -e
+      destination=&quot;%(TargetingPackSrcRequiringConstraintPatch.ILSrcFile)&quot;
```
dagood, line 69: This is different from `master`, and based on the variable name, I'd expect this new version to modify the Git-tracked source file rather than create an artifact. 😕 

https://github.com/dotnet/source-build-reference-packages/blob/a851216abf0f444ce15fcba1a3d5909e6a5ad914/src/targetPacks/Directory.Build.targets#L20

src/targetPacks/assemble.targets
```diff
@@ -1,19 +1,90 @@
...
+      <PatchedILSrcTemp>$(RepoRoot)artifacts/patchedILSrc/</PatchedILSrcTemp>
+    </PropertyGroup>
+    <ItemGroup>
+      <TargetingPackSrc
+        ILSrcFile="%(FullPath)"
+        RelativeOutputAssemblyFile="%(RecursiveDir)%(Filename).dll" />
+      <TargetingPackSrcRequiringConstraintPatch
+        ILSrcFile="$(PatchedILSrcTemp)$(MSBuildProjectName)/%(RecursiveDir)%(Filename).patched%(Extension)"
+        RelativeOutputAssemblyFile="%(RecursiveDir)%(Filename).dll"
+        DestinationFile="$(PatchedILSrcTemp)$(MSBuildProjectName)/%(RecursiveDir)%(Filename).patched%(Extension)" />
```
dagood, line 50: Is this attribute ever used in this new version of the patching target?

src/referencePackages/src/dir.props
dseefeld: I think 2.0.3 should work.  It is used in refeerence-package projects when including NETStandard.Library

src/targetPacks/assemble.targets
dseefeld: Fixed

src/targetPacks/assemble.targets
dseefeld: I updated to use this when transforming IL.

src/targetPacks/assemble.targets
dseefeld: I changed to use these names.

src/targetPacks/assemble.targets
dseefeld: Fixed.

src/targetPacks/assemble.targets
```diff
@@ -1,9 +1,81 @@
...
+    <Exec Command="
+      set -e
+      destination=&quot;%(TargetingPackSrcRequiringConstraintPatch.DestinationFile)&quot;
+      mkdir -p &quot;${destination%/*}&quot;
+      sed -E &quot;s/ constraint(([,)])|%24)/ 'constraint'\2/g&quot; &quot;%(TargetingPackSrcRequiringConstraintPatch.FullPath)&quot; > &quot;$destination&quot;" />
+
+  </Target>
+
+  <Target Name="AssembleTargetPack"
+          Condition="'@(TargetingPackSrcNotRequiringConstraintPatch)' != '' and '$(SkipTargetingPacks)' != 'true' ">
```
dagood, line 77: Shouldn't this still run even if all IL files require the constraint patch?

```suggestion
          Condition="'@(TargetingPackSrc)' != '' and '$(SkipTargetingPacks)' != 'true' ">
```

src/targetPacks/assemble.targets
```diff
@@ -1,9 +1,81 @@
 <?xml version="1.0" encoding="utf-8"?>
-<Project ToolsVersion="15.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+<Project DefaultTargets="AssembleTargetPack" ToolsVersion="15.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
```
dagood, line 2: IMO this is difficult to find and will be confusing in the future if it ends up combining with some other issue. I'd much rather add `Targets="AssembleTargetPack"` to the MSBuild call to be more explicit about what's intended to happen:

https://github.com/dotnet/source-build-reference-packages/blob/fe6afa24c69cc135200bbe1f12a90a11a450dcb0/src/targetPacks/buildTargettingPackages.proj#L34-L36

src/targetPacks/assemble.targets
dseefeld: Yes.  I fixed this. Thanks.

src/targetPacks/assemble.targets
dseefeld: Makes sense.  I updated as suggested.

dagood (approved):

PR #240, opened 2021-09-23
Port textOnly packages from pre-arcade to main

dseefeld:


MichaelSimons (approved):
Difficult to review given the size.  Since this is a merge between branches with autogenerated projects I am fine with merging.

PR #246, opened 2021-10-05
Remove old 1.0.0-alpha-5 ReferenceAssemblies

dseefeld:
This removes all 1.0.0-alpha-5 ReferenceAssemblies, since the 1.0.2 version has been added.   This will reduce the amount of code needed for SBRP in the tarball, since each version of the reference assembly code is quite sizable.

Also, changes any references in SBRP to 1.0.2.


dseefeld:
Closing.  Too large.  I will break this up.

PR #269, opened 2021-10-14
Add microsoft.extensions.commandlineutils.sources text-only package

MichaelSimons:
Fixes https://github.com/dotnet/source-build/issues/2511


src/textOnlyPackages/src/microsoft.extensions.commandlineutils.sources/5.0.1-rtm.21063.7/microsoft.extensions.commandlineutils.sources.csproj
```diff
@@ -0,0 +1 @@
+<Project Sdk="Microsoft.NET.Sdk" />
```
dseefeld, line 1: The project name should have the version number in it in case we need to differentiate it from other versions (not likely, but it is the pattern).

dseefeld (approved):
Just one comment.

src/textOnlyPackages/src/microsoft.extensions.commandlineutils.sources/5.0.1-rtm.21063.7/microsoft.extensions.commandlineutils.sources.csproj
MichaelSimons: Fixed

PR #270, opened 2021-10-15
Add prebuilt tracking for repo source-build builds

dseefeld:
Turning on prebuilt tracking for source-build builds.  This includes updating the prebuilt baseline with prebuilts that we know exist in SBRP.  Removal of these is tracked by https://github.com/dotnet/source-build/issues/2423.  

Also, adding a mechanism to add packages with dependent packages to SBRP.  This allows building the dependency packages first so they are available to the dependent projects without being reported as prebuilts.

### DIFF
--- a/eng/Build.props
+++ b/eng/Build.props
@@ -1,6 +1,19 @@
 <Project>
 
   <ItemGroup>
+    <!--
+      The following DependencyPackageProjects are ones on which other packages depend.  Adding them
+      to this ItemGroup will ensure that they get built first and in order of inclusion.  Also,
+      packages included here will be added to the source-build package cache when building with
+      source-build to allow them to be considered in prebuilt reporting.
+
+      Format:
+      <DependencyPackageProjects Include="$(RepoRoot)src\referencePackages\src\**\Microsoft.Extensions.Options.5.0.0.csproj" />
+    -->
+    <DependencyPackageProjects Include="$(RepoRoot)src\referencePackages\src\**\Microsoft.Extensions.Options.5.0.0.csproj" />
+    <DependencyPackageProjects Include="$(RepoRoot)src\referencePackages\src\**\Microsoft.Extensions.Logging.Abstractions.5.0.0.csproj" />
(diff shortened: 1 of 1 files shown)


eng/Build.props
```diff
@@ -1,16 +1,45 @@
 <Project>
 
   <ItemGroup>
+    <!--
+      The following DependencyPackageProjects are ones on which other packages depend.  Adding them
+      to this ItemGroup will ensure that they get built first and in order of inclusion.  Also,
```
dagood, line 6: This looks like https://github.com/dotnet/source-build-reference-packages/blob/pre-arcade/src/referencePackages/src/ProjectsToBuild.props, which I moved the repo away from with the migration to Arcade for a couple reasons:

* It's not great to maintain (this file will expand forever).
* It's fairly slow because it makes the build sequential rather than parallel.

I filed https://github.com/dotnet/source-build/issues/1690 at the time to track using `ProjectReferences` or something similar to handle building without relying on previously-source-built for every dependency.

Figuring out `ProjectReference` might not be feasible, but I think there's a decent middle ground: the dependency could be specified by adding an Item to a csproj that has a dependency, pointing at its dependency csproj. That way, the graph linkage isn't lost to a topological sort, and the info localized to the project that expresses the dependency.

dseefeld: The idea is that the projects included here are just for cases when packages + dependencies need to be added in servicing releases or PRs.  Once the packages make it to prev-source-built (i.e. a new release goes out, these can be removed).  I don't expect to be adding all project->project references here and I don't expect this to keep growing.  I specifically didn't want to define them in the project so I could define them all in one place for easier maintenance.

dagood: Oh! That makes sense. Just needs to be a part of the checklist then, I guess.

dseefeld: I agree.  Filed https://github.com/dotnet/source-build/issues/2522

dagood (approved):

PR #288, opened 2021-10-21
update Microsoft.Private.Intellisense version to latest

MichaelSimons:
Related to https://github.com/dotnet/source-build/issues/2551


src/textOnlyPackages/src/microsoft.private.intellisense/6.0.0-preview-20211019.1/IntellisenseFiles/net/1033/System.Text.Json.SourceGeneration.xml
```diff
@@ -0,0 +1,17 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<doc>
+  <assembly>
+    <name>System.Text.Json.SourceGeneration</name>
```
eerhardt, line 4: @carlossanlop @safern @gewarren - I thought we were going to filter this assembly from the drop. Why is it showing up in the Intellisense docs again?

gewarren: It's probably still there because we didn't update the APIs for RC 2. @safern Will that assembly be removed for GA?

safern: Yes, I will remove it for GA

eerhardt: Note that in 7 we already have 2 new source generators checked in: RegularExpressions and InteropServices. So this is going to be an ongoing thing we will want to solve.

PR #290, opened 2021-10-25
Update NETStandard.Library,2.0.3 to generate with 5.0 ILDasm

dseefeld:
Fixes: https://github.com/dotnet/source-build/issues/2567


MichaelSimons (approved):
Approved with the caveat that reviewing the netstandard.il changes are a little overwhelming and am trusting the tooling used to generate it.

PR #291, opened 2021-10-25
Add MS.DotNet.Common.ProjectTemplates.2.1,1.0.2-beta3

dseefeld:
Fixes: https://github.com/dotnet/source-build/issues/2560


MichaelSimons (commented):
How come this single 2.1 Common.ProjectTemplates package is added to SBRP but not the other 2.1 ProjectTemplates or the other out of support versions of ProjectTemplates are not?  I am trying to understand the rules/policies we should be using.  IMO if would be nice to eliminate the text only packages in installer.  They should either be moved to SBRP or the source repos should be added to source-build.

dseefeld:
> How come this single 2.1 Common.ProjectTemplates package is added to SBRP but not the other 2.1 ProjectTemplates or the other out of support versions of ProjectTemplates are not? I am trying to understand the rules/policies we should be using. IMO if would be nice to eliminate the text only packages in installer. They should either be moved to SBRP or the source repos should be added to source-build.

The difference is that this package is older and doesn't have an updated license.  Since the version number is not dynamic and the license needs to be updated, I added it to SBRP.

MichaelSimons (approved):

PR #294, opened 2021-12-03
Port referencePackage generation code from internal repo to SBRP

dseefeld:
Move package generation code from internal to SBRP.  This includes adding the code under `src/referencePackageSourceGenerator` and then setting it up to use the version of Arcade used in SBRP as well as porting it to use a 6.0 version of the SDK.  Multiple scripts from the internal repo were combined into one `generate.sh` script which includes generating a single package from the command line or multiple packages from a csv file as well as copying the resultant generated code to the current repo by default or another repo if specified.


generate.sh
```diff
@@ -0,0 +1,148 @@
...
+
+# Build the projects to generate source and projects
+"$scriptroot/eng/common/build.sh" --build --restore -bl /p:GeneratePackageSource=true /p:GeneratorVersion=2 $@
+
+# Copy source to destination
+pathToGeneratedSrc="$scriptroot/artifacts/generatedSrc"
+pathToRepoSrc="$pathToDestRepo/src/referencePackages/src"
+
+pushd "$pathToGeneratedSrc"
+cp ./ProjectsToBuildAdditions.txt ./CopyProjects.sh
```
MichaelSimons, line 138: What is `ProjectsToBuildAdditions.txt` all used for?  This is the only reference I see.  Would it make sense to change the format to be a simple project file list?  It would be more consumable here.

src/referencePackageSourceGenerator/Directory.Build.props
```diff
@@ -0,0 +1,49 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project>
+  <PropertyGroup>
+    <ImportNetSdkFromRepoToolset>false</ImportNetSdkFromRepoToolset>
+    <!-- <ExcludeRestorePackageImports>true</ExcludeRestorePackageImports> -->
```
MichaelSimons, line 5: What's the value in this commented out property?

generate.sh
```diff
@@ -0,0 +1,148 @@
...
+    pathToCSV="$defaultPathToCSV"
+fi
+
+# Generate restore projects for each target package
+# Packges are restored in their own project to eliminate any conflicts
+# between different versions of the same package.
+INPUT="$pathToCSV"
+OLDIFS=$IFS
+IFS=,
+DEFAULT_TFM=net6.0
```
MichaelSimons, line 110: Consider moving this up to the top as it is something that will need to change between branches.

MichaelSimons (commented):
I didn't review the tooling being ported closely.

generate.sh
dseefeld: In the pre-arcade version of this generation logic, the list of projects to build is manually maintained, so the instructions direct the user to copy from this file and paste into the projects to build props file.  This isn't needed in 6.0 and main, so I modified to just generate the copy script directly.

src/referencePackageSourceGenerator/Directory.Build.props
dseefeld: This is not needed.  I removed it.

generate.sh
dseefeld: Fixed.

MichaelSimons (approved):

PR #295, opened 2021-12-06
Removed unused packages

MichaelSimons:
The packages being removed here are not referenced by .NET 6.0.  These are packages that were once referenced but no longer.  Usually this is a result of upgrading to a newer version.


MichaelSimons:
Failing offline source-build leg is the result of a prebuilt tracked with https://github.com/dotnet/source-build/issues/2360

PR #298, opened 2021-12-07
Add Wcwidth.Sources,0.5.0 as text only package

dseefeld:
Fixes: https://github.com/dotnet/source-build/issues/2360


MichaelSimons:
This needs to be ported to release/6.0.

PR #301, opened 2021-12-07
Add CODEOWNERS file

MichaelSimons:


CODEOWNERS
```diff
@@ -0,0 +1,4 @@
+# Users referenced in this file will automatically be requested as reviewers for PRs that modify the given paths.
+# See https://help.github.com/articles/about-code-owners/
+
+@dotnet/source-build-internal
```
lbussell, line 4: Should we include a wildcard file pattern here? [Documentation](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners#example-of-a-codeowners-file)

```suggestion
* @dotnet/source-build-internal
```

MichaelSimons: You are right, I misread the documentation.

lbussell (approved):

PR #335, opened 2022-02-16
Add System.Runtime.CompilerServices.Unsafe 6.0.0 and System.Text.Encoding.CodePages 6.0.0

lbussell:
Eliminates the need for one of the patches in https://github.com/dotnet/roslyn/pull/58727


lbussell:
> LGTM, please update the PR description to remove the testing comment.

Done.

> Is this something that will get backported to 6.0?

It can be, it will remove the need for this patch: https://github.com/dotnet/installer/commit/374f8b5faa148c0303c3d0a5d5d75c168c3220ac

PR #357, opened 2022-03-11
Add Microsoft.CodeAnalysis.CSharp 4.0.1

chsienki:
Generated via `.\generate.sh --pkg Microsoft.CodeAnalysis.CSharp,4.0.1` 

Needed so we can pin the razor generator to version 4.0.1 in the SDK without diverging source build (see https://github.com/dotnet/sdk/pull/24130 )

I think it looks right, but I'm new at this :)


chsienki:
@MichaelSimons I've removed the `netcoreapp3.1` versions as they don't seem to be supported, and manually fixed up the missing internal refs that was causing it to fail.

Is that an acceptable way to resolve?

MichaelSimons:
@chsienki - Can you explain in more detail what you meant by "netcoreapp3.1 versions as they don't seem to be supported"?  I looked at the first commit's build failures and noticed an issue with the tooling that I just submitted a [PR ](https://github.com/dotnet/source-build-reference-packages/pull/359) for.  

I will regenerate the source with this fix and update this PR.  Aside from that, there do appear to be a number of issues in the generated source that will need to be analyzed and corrected.

MichaelSimons:
@chsienki - I see you made a number of hand edits already.  Do you know if these are something we can tool?  GenAPI is the underlying tooling being used.

MichaelSimons:
I regenerated the source but did not re-add microsoft.codeanalysis.common and microsoft.codeanalysis.csharp 3.1 nor did I undo the hand edits.

Ideally the hand edits would be minimal so I am looking for input on how to generate these ref packages correctly.  Having to make these many manual edits doesn't scale.  Also we would ideally not be removing 3.1.

chsienki:
@MichaelSimons Ok, I'll see if I can update the tool to generate the correct files.

src/referencePackages/src/microsoft.codeanalysis.common/4.0.1/lib/netstandard2.0/Microsoft.CodeAnalysis.cs
```diff
@@ -0,0 +1,5047 @@
...
+    }
+    public enum VarianceKind : short
+    {
+        None = (short)0,
+        Out = (short)1,
+        In = (short)2,
+    }
+    public static partial class WellKnownDiagnosticTags
+    {
+        public const string AnalyzerException
```
MichaelSimons, line 4994: Consider pulling these into a separate file so that we can re-run the generator.  It will also help callout the manual fixes.

MichaelSimons (commented):
Can you rerun the generator to get the up-to-date changes?

MichaelSimons:
This PR has been superseded by https://github.com/dotnet/source-build-reference-packages/pull/416.

PR #359, opened 2022-03-15
Fix concurrency issue in the referencePackageSourceGenerator

MichaelSimons:
The logic that replaces the assembly attributes placeholder in GenerateProjects was running before GenerateSource completed.  This causes the generated source to be incorrect.

This was encountered with https://github.com/dotnet/source-build-reference-packages/pull/357.

### DIFF
--- a/src/referencePackageSourceGenerator/GenerateProjects/GenerateProjects.proj
+++ b/src/referencePackageSourceGenerator/GenerateProjects/GenerateProjects.proj
@@ -18,7 +18,7 @@
   <UsingTask AssemblyFile="$(GenerateProjectsTaskPath)" TaskName="GenerateProjects" />
 
   <Target Name="CoreCompile" DependsOnTargets="GetGenApiDlls" >
-    <!-- Ensure source is generated first.  GenerateProjects modifies updates the assembly version attribute placeholder -->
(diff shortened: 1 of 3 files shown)


src/referencePackageSourceGenerator/GenerateProjects/GenerateProjects.proj
```diff
@@ -18,6 +18,9 @@
   <UsingTask AssemblyFile="$(GenerateProjectsTaskPath)" TaskName="GenerateProjects" />
 
   <Target Name="CoreCompile" DependsOnTargets="GetGenApiDlls" >
+    <!-- Ensure source is generated first.  GenerateProjects modifies updates the assembly version attribute placeholder -->
+    <MSBuild Projects="@(ProjectReference)" Targets="Build" />
```
lbussell, line 22: Is "modifies updates" a typo?

lbussell (approved):
Thanks for diagnosing this. I've hit this before too but didn't know what to investigate/thought it was user error.

src/referencePackageSourceGenerator/GenerateProjects/GenerateProjects.proj
MichaelSimons: Yes it is - fixed.

crummel (approved):

PR #364, opened 2022-03-18
Update the version of GenAPI used and pass through correct arguments

chsienki:


MichaelSimons (commented):
The new GenAPI version utilizes c# 8.0 features which is not compatible with the generated projects.  Can you please update the [project template](https://github.com/dotnet/source-build-reference-packages/blob/main/src/referencePackageSourceGenerator/GenerateProjects/PackageProjectTemplate.xml) accordingly?

chsienki:
@MichaelSimons Good catch, I pulled this out from another branch and missed that. 

The newer GenAPI tool understands nullable and generates with the annotations. I chose not to enable nullability in the generated projects and just suppress the warnings about using it in a non-null context (`CS8632`) as there seemed to be a lot of non trivial warnings added when enabling it (even after suppressing `CS8597` for `throw null`).

Let me know if you'd rather switch that around.

src/referencePackageSourceGenerator/GenerateProjects/PackageProjectTemplate.xml
```diff
@@ -3,6 +3,8 @@
 
   <PropertyGroup>
     <TargetFrameworks>#TargetFrameworks#</TargetFrameworks>
+	<LangVersion>latest</LangVersion>
+	<NoWarn>8632</NoWarn>
```
MichaelSimons, line 7: This should include $(NoWarn).  Please moved to https://github.com/dotnet/source-build-reference-packages/blob/main/src/referencePackages/Directory.Build.props#L47 so that all warning suppressions are centrally located.  A comment should be added like the other NoWarns indicating the reason.

Without this the generated code has numerous warnings.

```suggestion
	<NoWarn>$(NoWarn);8632</NoWarn>
```

src/referencePackageSourceGenerator/GenerateProjects/PackageProjectTemplate.xml
```diff
@@ -3,6 +3,8 @@
 
   <PropertyGroup>
     <TargetFrameworks>#TargetFrameworks#</TargetFrameworks>
+	<LangVersion>latest</LangVersion>
```
MichaelSimons, line 6: Rethinking about the patterns used in this repo, I think this may be better defined in the common props - https://github.com/dotnet/source-build-reference-packages/blob/main/src/referencePackages/Directory.Build.props.  This gives us more flexibility without having to edit/regen all projects.  The projects should just contain specializations.

MichaelSimons (changes requested):

MichaelSimons (approved):
Thanks for the tooling improvements.

PR #376, opened 2022-05-04
[main] Bump System.Security.Cryptography.X509Certificates 4.1.0 to 4.1.2

lbussell:
Fixes https://github.com/dotnet/source-build/issues/2853


MichaelSimons (commented):
A x509Cert.5.0.0 dependency should be added to the eng/build.props file.  Without this, the build may fail intermittently.

crummel (approved):

MichaelSimons (approved):

PR #395, opened 2022-07-05
Add text-only packages required for 7.0

MichaelSimons:
This change adds the following text-only packages:

Wcwidth.Sources 0.6.0
Microsoft.Build.NoTargets 3.5.0
Microsoft.Build.Traversal 3.1.6
~~Microsoft.DotNet.CilStrip.Sources 7.0.0-beta.22313.1~~

This change removs the following text-only packages:
Wcwidth.Sources 0.5.0
Microsoft.Build.NoTargets 3.1.0
Microsoft.Build.Traversal 3.0.23


lbussell:
Are we sure that CilStrip.Sources doesn't need to be generated at tarball creation time? That version number seems like it's in development and may be updated when net7.0 releases.

Alternatively, will this prebuilt go away with this issue being resolved? It seems like the preferable approach to me, unless I'm missing something. This is from a previous investigation we did, I think. https://github.com/dotnet/runtime/issues/69460

MichaelSimons:
> This is from a previous investigation we did, I think. https://github.com/dotnet/runtime/issues/69460

You are correct.  This should be resolved by runtime, not SBRP.  I will remove this package.

lbussell (approved):

PR #396, opened 2022-07-05
Remove package unreferenced by .NET 7.0

MichaelSimons:
Removed the following packages:

microsoft.aspnetcore.app.ref.5.0.0.nupkg
microsoft.aspnetcore.components.webassembly.templates.3.2.1.nupkg
microsoft.dotnet.platformabstractions.2.1.0.nupkg
microsoft.net.sdk.il.6.0.0-rc.1.21415.6.nupkg
microsoft.private.intellisense.6.0.0-preview-20211019.1.nupkg
nunit3.dotnetnew.template.1.5.3.nupkg
nunit3.dotnetnew.template.1.6.5.nupkg
nunit3.dotnetnew.template.1.7.2.nupkg
nunit3.dotnetnew.template.1.8.1.nupkg
system.diagnostics.diagnosticsource.4.0.0.nupkg
system.diagnostics.diagnosticsource.4.4.0.nupkg
system.diagnostics.stacktrace.4.3.0.nupkg
system.globalization.calendars.4.0.1.nupkg
system.globalization.extensions.4.0.1.nupkg
system.io.compression.4.1.0.nupkg
system.io.memorymappedfiles.4.3.0.nupkg
system.io.pipes.accesscontrol.4.5.1.nupkg
system.net.requests.4.0.11.nupkg
system.security.cryptography.cng.4.2.0.nupkg
system.security.cryptography.csp.4.0.0.nupkg
system.security.cryptography.openssl.4.0.0.nupkg
system.security.cryptography.x509certificates.4.1.2.nupkg
system.security.principal.windows.4.0.0.nupkg
system.text.json.5.0.2.nupkg
system.valuetuple.4.4.0.nupkg

PR #402, opened 2022-07-13
Bump Microsoft.CodeAnalysis.Collections to 4.2.0-1.22102.8

lbussell:
Microsoft.CA.Collections was bumped from 4.0.0-4.21379.20 to 4.2.0-1.22102.8 in msbuild, so we need to bump the version here in SBRP too to remove the prebuilt.


MichaelSimons (approved):

lbussell:
> Was this a recent bump? Is msbuild the only repo to reference this? I ask because I didn't find the old package as unreferenced in the cleanup work I recently did in #396.

Per our discussion offline, I added the old package back for now just in case. if it's truly not needed, we can clean it up later.

crummel (approved):

MichaelSimons (approved):

PR #423, opened 2022-08-08
Remove unreferenced packages

MichaelSimons:
Removing the following packages that do not appear to be referenced in a tarball build:
system.security.cryptography.primitives.4.0.0
system.security.cryptography.protecteddata.4.7.0


lbussell:
It looks like you also removed System.Security.Cryptography.Primitives **4.3.0** as well, was this intentional?

MichaelSimons:
> It looks like you also removed System.Security.Cryptography.Primitives **4.3.0** as well, was this intentional?

That was a mistake on my part - added back.

crummel (approved):

lbussell (approved):

PR #431, opened 2022-08-25
Add netcore.app 2.0.3 target pack graph

### LINKED ISSUE #650: Xliff-Tasks downloads 2.0.X prebuilt packages during source-build
The following packages are reported as source-build prebuilts for 7.0rc1:

- Microsoft.NETCore.App.2.0.3
- Microsoft.NETCore.DotNetAppHost.2.0.3
- Microsoft.NETCore.DotNetHostPolicy.2.0.3
- Microsoft.NETCore.DotNetHostResolver.2.0.3
- Microsoft.NETCore.Platforms.2.0.1

These packages should not be referenced or restored during the xliff-tasks build, as they are not available during source-build.

Steps to reproduce:

1. clone xliff-tasks
2. Run `./build.sh -c Release --restore --build --pack /p:ArcadeBuildFromSource=true -bl`
3. Run the following commands to see the restored prebuilts (the versions are slightly different than what we reported in the end-to-end source-build):
    `find artifacts/source-build/self/package-cache/ -iname *2.2.8*`
    `find artifacts/source-build/self/package-cache/ -iname *2.0.*.nupkg`

https://github.com/dotnet/installer/pull/14372 is a workaround, but including the prebuilts is only acceptable for RC1.

MichaelSimons:
Fixes https://github.com/dotnet/xliff-tasks/issues/650


lbussell:
I see xliff-tasks/main is targeting 2.2.3 now. Should we add this as well? I would expect it to flow into 7.0 (either automatically or manually) since it came before the net7.0 TFM change in that repo.

https://github.com/dotnet/xliff-tasks/blob/e0068bf9c8bf13d84d77f19a2628e910bdabdae7/src/XliffTasks/XliffTasks.csproj#L32

lbussell (approved):
I don't have issues with the content of this PR, I think it's just a question of whether we should have 2.2.3 or let xliff-tasks roll back to 2.0.3. I don't know which is more correct, but since this PR looks OK, I will let you decide whether to merge or not.

MichaelSimons:
Good catch @lbussell, I asked about this in https://github.com/dotnet/xliff-tasks/issues/650.

MichaelSimons:
Closing as the xliff change to introduce the prebuilts was reverted.

PR #443, opened 2022-09-16
Add csproj for new ASP.NET ref pack.

crummel:


lbussell (approved):
Tested locally and the package `Microsoft.AspNetCore.App.Ref.6.0.0.nupkg` is produced now.

PR #455, opened 2022-09-30
[release/7.0] Add Microsoft.CodeAnalysis 3.7.0 and 3.9.0

lbussell:
I mostly followed Michael's steps here: https://github.com/dotnet/source-build-reference-packages/pull/416

For 3.9.0 I separated the generated code from my tweaks so it is easier to see what I had to change.

I am still testing this in a tarball and still need to verify some things in IL - I'll mark as ready to review once I'm done with those things.


MichaelSimons:
@lbussell - Is this PR still applicable?

lbussell:
If roslyn-analyzers decides to update their version of Microsoft.CodeAnalysis in the future, these could be useful for the direct references to MS.CodeAnalysis 3.9.0 that remain in the code. See https://github.com/dotnet/roslyn-analyzers/pull/6196 for more context.

I'll close for now but leave the branch so it can be re-opened or referenced in the future.

PR #478, opened 2022-11-13
[xml] System.Net.Sockets.Socket.Close, fix wrong timeout time unit

viordash:


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/source-build-reference-packages?pullRequest=478) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dotnet/source-build-reference-packages?pullRequest=478) before we can accept your contribution.<br/><br/>:x: viordash [sign now](https://cla.dotnetfoundation.org/dotnet/source-build-reference-packages?pullRequest=478)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dotnet/source-build-reference-packages?pullRequest=478) it.</sub>

MichaelSimons:
@viordash - Can you please provide some background on why this change is needed?  TIA

viordash:
I got an error when trying to send any remaining data because I set a timeout for Socket.Close in seconds based on a method hint in VS2022 (.NET6).

Then checked VS2019 (.NET Core 3.1) and the Socket.Close method hint also has this message:
`timeout: Wait up to timeout seconds to send any remaining data, then close the socket.`
But in the [help for this method](https://learn.microsoft.com/en-us/dotnet/api/system.net.sockets.socket.close?view=net-7.0#system-net-sockets-socket-close (system-int32)) says that the timeout is in milliseconds.

To be sure, I checked the [source code](https://github.com/dotnet/runtime/blob/8918adf2e5255920aed3c61f9b92c2a51eec0fe3/src/libraries/System.Net.Sockets/src/System/Net/Sockets/Socket.cs#L3209) and made sure that milliseconds is the correct unit of time.

NET Framework 4.8 (and probably older versions too) also has the wrong unit of time, but I don't know where to fix it.

MichaelSimons:
@viordash - thanks for the backstory.  This is not correct place to fix this issue.  I would suggest logging a [dotnet/runtime](https://github.com/dotnet/runtime) issue regarding the intellisense documentation being incorrect.  This repo is a mirror of sorts of the documentation you want updated; it is not the origin of the source.  This repo contains the reference packages required to be able to source-build .NET.

viordash:
Thanks for the right way. 
I made a [issue](https://github.com/dotnet/runtime/issues/78348)

PR #484, opened 2022-11-28
Add the Roslyn-based GenAPI as a secondary backend. Introduce --genapi-backend option.

andriipatsula:
* Add `dotnet8-transport` to a list of package sources
* Add `--genapi-backend` option [cci|roslyn] with the default value `cci`
* Based on parameters either the Roslyn-based or the Cci-based GenAPI is used to generate reference assemblies.

https://github.com/dotnet/arcade/issues/11726

p.s. The Roslyn-based GenAPI (based on SyntaxGenerator and CSharpSyntaxRewriter) is in testing phase. I created this PR to communicate the changes and usage of the tooling. We can merge it or wait till the Testing is finished.

### DIFF
--- a/generate.sh
+++ b/generate.sh
@@ -12,6 +12,8 @@ scriptroot="$( cd -P "$( dirname "$source" )" && pwd )"
 RED='\033[0;31m'
 NC='\033[0m'
 
+backend="cci"
+
 usage() {
     echo "usage: $0 [options]"
     echo ""
@@ -35,6 +37,7 @@ usage() {
     echo "  --dest <pathToDestRepo>            A path to the root of the repo to copy source into."
     echo "  --type <packageType>               Type of the package to generate. Accepted values: ref (default) | text."
     echo "  --feeds <nugetFeeds>               A semicolon-separated list of additional NuGet feeds to use during restore."
+    echo "  --genapi-backend                   The GenAPI backend used to generate reference assemblies [cci|roslyn]. Default is '$backend'."
     echo ""
 }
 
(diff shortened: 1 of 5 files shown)


generate.sh
```diff
@@ -35,6 +37,7 @@ usage() {
     echo "  --dest <pathToDestRepo>            A path to the root of the repo to copy source into."
     echo "  --type <packageType>               Type of the package to generate. Accepted values: ref (default) | text."
     echo "  --feeds <nugetFeeds>               A semicolon-separated list of additional NuGet feeds to use during restore."
+    echo "  --genapi-backend                   The GenAPI backend used to generate reference assemblies [cci|roslyn]. Default is '$backend'."
```
MichaelSimons, line 40: The pattern used to indicate the accepted and default values is inconsistent with the `type` option.  The script should use a consistent pattern.  I personally don't care which pattern is used.

generate.sh
```diff
@@ -12,6 +12,8 @@ scriptroot="$( cd -P "$( dirname "$source" )" && pwd )"
 RED='\033[0;31m'
 NC='\033[0m'
 
+backend="cci"
```
MichaelSimons, line 15: I feel like it would be best to group the option parameters together for readability/maintainability.  Consider pulling the others up here.

MichaelSimons (commented):
I think we should merge this ahead of testing being completed.  I do however suggest we change the default to cci until the testing is finished.

generate.sh
andriipatsula: Updated the description of the `--genapi-backend` option

MichaelSimons (approved):

PR #498, opened 2022-12-15
Enable CodeQL

crummel:


azure-pipelines/builds/ci.yml
```diff
@@ -35,3 +35,7 @@ stages:
       #Nuget and Signing Validation are not needed as we only produce a transport package.
       enableSigningValidation: false
       enableNugetValidation: false
+
+variables:
```
MichaelSimons, line 39: In general, I find it best to put the stages at the end of the yml.  The other properties should come first.  The primary reason for this is the stage(s) section can be very long and the variables can get overlooked.  Example: https://github.com/dotnet/installer/blob/main/.vsts-ci.yml#L26.  While it is not too much of an issue with this simple pipeline, I think it is best to follow this formatting style.

crummel: Thanks, changed as suggested.

MichaelSimons (approved):

PR #506, opened 2023-02-01
add System.Reflection.Metadata 6.0.1

oleksandr-didyk:
Done as part of https://github.com/dotnet/runtime/issues/81468

`runtime` is manually enforcing the patched 6.0.1 version through [`Versions.props`](https://github.com/dotnet/runtime/blob/main/eng/Versions.props#L130) as part of https://github.com/dotnet/runtime/pull/73871

### DIFF
--- a/src/referencePackages/src/system.buffers/4.5.1/System.Buffers.4.5.1.csproj
+++ b/src/referencePackages/src/system.buffers/4.5.1/System.Buffers.4.5.1.csproj
@@ -1,40 +1,44 @@
 <Project Sdk="Microsoft.NET.Sdk">
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.props))/dir.props" />
+
   <PropertyGroup>
     <TargetFrameworks>netstandard1.1;netstandard2.0;net45;net461</TargetFrameworks>
     <NuspecFile>$(ArtifactsBinDir)system.buffers/4.5.1/system.buffers.nuspec</NuspecFile>
     <StrongNameKeyId>Open</StrongNameKeyId>
-   </PropertyGroup>
+  </PropertyGroup>
 
   <PropertyGroup>
     <OutputPath>$(ArtifactsBinDir)system.buffers/4.5.1/ref/</OutputPath>
     <IntermediateOutputPath>$(ArtifactsObjDir)system.buffers/4.5.1</IntermediateOutputPath>
   </PropertyGroup>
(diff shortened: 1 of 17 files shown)


MichaelSimons:
General comments:

1. When adding new packages, I typically revert any edits to existing packages to reduce the burden on folks reviewing the changes.  It is best to do this clean up either in a separate PR or at a minimum a separate commit.
2. It's a good idea to validate changes locally prior to opening a PR in this repo.  The build is fairly quick.  Your changes fail to build because of https://github.com/dotnet/source-build/issues/1957.
3. I am concerned this package is going to get accidentally cleaned up next time we do a scan for unreferenced SBRPs.  The scan happens in the source-build context.  What is the change being made to runtime that will get it to use the SBRP in the source-build context versus picking up latest.  This is what is required for this package to not get cleaned up.

MichaelSimons:
I forgot one more thing:

4. Any new projects need to be added as a DependencyPackageProjects [here](https://github.com/dotnet/source-build-reference-packages/blob/main/eng/Build.props#L31)

oleksandr-didyk:
> General comments:
> 
> 1. When adding new packages, I typically revert any edits to existing packages to reduce the burden on folks reviewing the changes.  It is best to do this clean up either in a separate PR or at a minimum a separate commit.
> 2. It's a good idea to validate changes locally prior to opening a PR in this repo.  The build is fairly quick.  Your changes fail to build because of [[SBRP] generate.sh doesn't handle "<icon>" in reference package nuspecs source-build#1957](https://github.com/dotnet/source-build/issues/1957).
> 3. I am concerned this package is going to get accidentally cleaned up next time we do a scan for unreferenced SBRPs.  The scan happens in the source-build context.  What is the change being made to runtime that will get it to use the SBRP in the source-build context versus picking up latest.  This is what is required for this package to not get cleaned up.

1. Good point, thank you. Removed the unrelated changes.
2. That's on me, sorry. Removed the icon tag and re-ran the build locally, should be A-OK now
3. The package version added is already in use in runtime, meaning that once dependency on SBRP is added to runtime, its source-build should pick this version up. The SBRP  dependency should be added as part of https://github.com/dotnet/runtime/issues/81468 (If I understood correctly your point here)

> 4. Any new projects need to be added as a DependencyPackageProjects [here](https://github.com/dotnet/source-build-reference-packages/blob/main/eng/Build.props?rgh-link-date=2023-02-01T22%3A19%3A42Z#L31)

4. Thanks for the pointer, added

MichaelSimons (approved):

PR #514, opened 2023-02-07
remove icons from nuspec files

oleksandr-didyk:
Resolves https://github.com/dotnet/source-build/issues/1957

Tested by re-generating SBRPs for `System.Reflection.Metadata 6.0.1`, which would in the past alter the `*.nuspec` files to contain the `<index>` attribute. After the change the attribute is not present + the repo build is green


MichaelSimons (approved):
Thanks for fixing this!

PR #524, opened 2023-02-14
add support for multiple tfms in ref generation

oleksandr-didyk:
Trivial improvement to allow for multiple TFMs to be specified for package generation

### DIFF
--- a/generate.sh
+++ b/generate.sh
@@ -27,10 +27,9 @@ usage() {
     echo "  Either --pkg or --pkgCsv must be specified"
     echo ""
     echo "options:"
-    echo "  --pkg <packageName,version[,tfm]>  A package and version, no spaces, separated by comma.  If the optional"
-    echo "                                       tfm is specified, it will be used in the project that restores the"
-    echo "                                       package."
-    echo "                                       Example: System.Text.Json,4.7.0"
+    echo "  --pkg <packageName,version[,tfm]>  A package and version, no spaces, separated by comma.  If optional semicolon-separated"
+    echo "                                       TFMs are specified, they will be used in the project that restores the package."
+    echo "                                       Example: System.Text.Json,4.7.0,nestandard2.0;net461"
     echo "  --pkgCsv <pathToCSV>               A path to a csv file of packages to generate. Format is the same as the --pkg"
     echo "                                       option above, one per line.  If specified, the --pkg option is ignored."
     echo "  --dest <pathToDestRepo>            A path to the root of the repo to copy source into."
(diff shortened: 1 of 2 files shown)


src/referencePackageSourceGenerator/GenerateSource/targetPackage.csproj.template
```diff
@@ -2,7 +2,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>##Tfm##</TargetFramework>
+    <##TfmAttr##>##Tfm##</##TfmAttr##>
```
MichaelSimons, line 5: I find it a bit odd to refer to this as an attribute in the scope of an msbuild project defined in xml.  Consider naming it `TfmProperty` or `TfmPropertyName`

generate.sh
```diff
@@ -27,10 +27,9 @@ usage() {
     echo "  Either --pkg or --pkgCsv must be specified"
     echo ""
     echo "options:"
-    echo "  --pkg <packageName,version[,tfm]>  A package and version, no spaces, separated by comma.  If the optional"
-    echo "                                       tfm is specified, it will be used in the project that restores the"
-    echo "                                       package."
-    echo "                                       Example: System.Text.Json,4.7.0"
+    echo "  --pkg <packageName,version[,tfm]>  A package and version, no spaces, separated by comma.  If optional semicolon-separated"
```
MichaelSimons, line 30: Consider renaming `tfm` to `tfms` as it betters calls out the support for multiples.

generate.sh
```diff
@@ -27,10 +27,9 @@ usage() {
...
+    echo "                                       TFMs are specified, they will be used in the project that restores the package."
+    echo "                                       Example: System.Text.Json,4.7.0,nestandard2.0;net461"
```
MichaelSimons, line 32: Consider adding multiple examples to make it clear that the tfms are optional.
```suggestion
    echo "                                       Example: System.Collections,4.3.0"
    echo "                                       Example: System.Text.Json,4.7.0,nestandard2.0;net461"
```

src/referencePackageSourceGenerator/GenerateSource/targetPackage.csproj.template
oleksandr-didyk: Sure thing, renamed to `TfmPropertyName`

generate.sh
oleksandr-didyk: Good point, thanks. Renamed

generate.sh
oleksandr-didyk: Good suggestion, thanks. Added

MichaelSimons (approved):

PR #546, opened 2023-02-24
Add Microsoft.Extensions.ObjectPool 6.0.0

mthalman:
Adds Microsoft.Extensions.ObjectPool 6.0.0 due to prebuilt being introduced for this package from https://github.com/dotnet/razor. This looks to have been caused by https://github.com/dotnet/razor/pull/8289.

### DIFF
--- a/src/referencePackages/src/microsoft.extensions.objectpool/6.0.0/Microsoft.Extensions.ObjectPool.6.0.0.csproj
+++ b/src/referencePackages/src/microsoft.extensions.objectpool/6.0.0/Microsoft.Extensions.ObjectPool.6.0.0.csproj
@@ -1,0 +1,36 @@
+<Project Sdk="Microsoft.NET.Sdk">
+  <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.props))/dir.props" />
+
+  <PropertyGroup>
+    <TargetFrameworks>netstandard2.0;net461</TargetFrameworks>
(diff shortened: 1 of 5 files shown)


MichaelSimons (commented):
FYI, From https://github.com/dotnet/source-build-reference-packages#reference

After generating new reference packages, all new projects must be referenced as a [DependencyPackageProjects](https://github.com/dotnet/source-build-reference-packages/blob/main/eng/Build.props#L10). These must be defined in dependency order.

mthalman:
> FYI, From https://github.com/dotnet/source-build-reference-packages#reference
> 
> After generating new reference packages, all new projects must be referenced as a [DependencyPackageProjects](https://github.com/dotnet/source-build-reference-packages/blob/main/eng/Build.props#L10). These must be defined in dependency order.

This is what confuses me. Why is only one listed when there are a whole bunch of other packages in src?

MichaelSimons:
> FYI, From https://github.com/dotnet/source-build-reference-packages#reference
> 
> After generating new reference packages, all new projects must be referenced as a [DependencyPackageProjects](https://github.com/dotnet/source-build-reference-packages/blob/main/eng/Build.props#L10). These must be defined in dependency order.

Only new projects are added here until the previous source-build-artifacts get updated in which they are removed.  See for details: https://github.com/dotnet/source-build-reference-packages/blob/main/eng/Build.props#L11.  This all related to the problem of project references not being utilized in the projects.

MichaelSimons (approved):

PR #565, opened 2023-03-08
Add AssemblyKeyFile and AssemblyDelaySign attributes to the exclusion list.

andriipatsula:
The reference package for the System.Threading,4.0.11 package and the netstandard1.0 TFM contains two assembly attributes that produces errors:

>   CSC : error CS1616: Option 'PublicSign' overrides attribute '**System.Reflection.AssemblyDelaySignAttribute**' given in a source file or added module [/__w/1/s/artifacts/source-build/self/src/src/referencePackages/src/system.threading/4.0.11/System.Threading.4.0.11.csproj::TargetFramework=**netstandard1.0**] [/__w/1/s/.packages/microsoft.dotnet.arcade.sdk/8.0.0-beta.23153.1/tools/Build.proj]
  CSC : error CS1616: Option 'CryptoKeyFile' overrides attribute '**System.Reflection.AssemblyKeyFileAttribute**' given in a source file or added module [/__w/1/s/artifacts/source-build/self/src/src/referencePackages/src/system.threading/4.0.11/System.Threading.4.0.11.csproj::TargetFramework=**netstandard1.0**] [/__w/1/s/.packages/microsoft.dotnet.arcade.sdk/8.0.0-beta.23153.1/tools/Build.proj]

**AssemblyKeyFileAttribute** - specifies the name of a file containing the key pair used to generate a strong name. As the public key is not present in the reference package - it's OK to filter out such attribute.
**AssemblyDelaySignAttribute** - attribute is used on an assembly, space is reserved for the signature which is later filled by a signing tool such as the Sn.exe utility.

### DIFF
--- a/src/referencePackageSourceGenerator/DefaultGenApiDocIds.txt
+++ b/src/referencePackageSourceGenerator/DefaultGenApiDocIds.txt
@@ -48,3 +48,5 @@ T:System.Xml.Serialization.XmlEnumAttribute
 T:System.Xml.Serialization.XmlIgnoreAttribute
 T:System.Xml.Serialization.XmlRootAttribute
 T:System.Windows.Markup.ContentPropertyAttribute
+T:System.Reflection.AssemblyKeyFileAttribute
+T:System.Reflection.AssemblyDelaySignAttribute


andriipatsula:
The problem appears in the issue: https://github.com/dotnet/sdk/issues/31006

/cc @ViktorHofer

ViktorHofer:
Do we have any reports about the issue happening on other TFMs and/or packages as well?

andriipatsula:
> Do we have any reports of the issue happening on other TFMs and/or packages as well?

Nope, there were no such issues for other TFMs (package: System.Threading,4.0.11). I don't have enough data to make a conclusion about how often this problem will occur as we just introduced the assembly reference generation by the GenAPI.

I don't think we must worry about these two attributes as, previously, the SBRP generator never includes them in the reference packages.

src/referencePackageSourceGenerator/DefaultGenApiDocIds.txt
```diff
@@ -48,3 +48,5 @@ T:System.Xml.Serialization.XmlEnumAttribute
 T:System.Xml.Serialization.XmlIgnoreAttribute
 T:System.Xml.Serialization.XmlRootAttribute
 T:System.Windows.Markup.ContentPropertyAttribute
+T:System.Reflection.AssemblyKeyFileAttribute
```
MichaelSimons, line 51: Consider alphabetizing to improve the readability.

crummel (approved):

PR #568, opened 2023-03-09
Add System.Text.Json 6.0.0 and dependencies

mthalman:
Adds the following packages as SBRPs:
* System.Runtime.CompilerServices.Unsafe.6.0.0
* Microsoft.Bcl.AsyncInterfaces.6.0.0
* System.Text.Encodings.Web.6.0.0
* System.Text.Json.6.0.0

These are needed due to the prebuilts being introduced by https://github.com/dotnet/source-build-externals/pull/112

### DIFF
--- a/src/referencePackages/src/system.text.json/6.0.0/lib/net6.0/System.Text.Json.cs
+++ b/src/referencePackages/src/system.text.json/6.0.0/lib/net6.0/System.Text.Json.cs
@@ -1,0 +1,1794 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+// ------------------------------------------------------------------------------
+// Changes to this file must follow the http://aka.ms/api-review process.
+// ------------------------------------------------------------------------------
+[assembly: System.Runtime.CompilerServices.CompilationRelaxations(8)]
+[assembly: System.Runtime.CompilerServices.RuntimeCompatibility(WrapNonExceptionThrows = true)]
+[assembly: System.Diagnostics.Debuggable(System.Diagnostics.DebuggableAttribute.DebuggingModes.IgnoreSymbolStoreSequencePoints)]
+// Commenting out because JsonSerializerOptionsUpdateHandler is an internal type
+//[assembly: System.Reflection.Metadata.MetadataUpdateHandler(typeof(JsonSerializerOptionsUpdateHandler))]
+[assembly: System.Runtime.Versioning.TargetFramework(".NETCoreApp,Version=v6.0", FrameworkDisplayName = "")]
+[assembly: System.CLSCompliant(true)]
+[assembly: System.Reflection.AssemblyDefaultAlias("System.Text.Json")]
+[assembly: System.Resources.NeutralResourcesLanguage("en-US")]
+[assembly: System.Reflection.AssemblyMetadata(".NETFrameworkAssembly", "")]
+[assembly: System.Reflection.AssemblyMetadata("Serviceable", "True")]
+[assembly: System.Reflection.AssemblyMetadata("PreferInbox", "True")]
+[assembly: System.Reflection.AssemblyMetadata("IsTrimmable", "True")]
+[assembly: System.Reflection.AssemblyCompany("Microsoft Corporation")]
+[assembly: System.Reflection.AssemblyCopyright("© Microsoft Corporation. All rights reserved.")]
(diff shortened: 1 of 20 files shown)


MichaelSimons (approved):

MichaelSimons:
Looks like the source-built targeting packs are not getting loaded therefore there are prebuilts.  One issue that started with 6.0 is that with the inclusion of the analyzers/generators, the targeting pack gets serviced and the SDK is always trying to load the most recent version.  The SBRP targeting packs don't get regenerated for each servicing release.  Given that they don't contain the analyzers/generators this doesn't feel necessary for the intended usage.  This is the downside of utilizing the implicit references.

cc @ViktorHofer

ViktorHofer:
We should fix that by modifying the KnownFrameworkReference item metadata to point to the package version that source build produces instead of what the SDK decides to target. We do something similar in dotnet/runtime to make sure that the live built asset is chosen.

The SDK has known framework references which are encoded in the SDK layout: `"sdk\<sdkverison>\Microsoft.NETCoreSdk.BundledVersions.props"`

image

We want to change the `TargetingPackVersion` item metadata from 6.0.11 to 6.0.0 in the example of net6.0. This can be achieved via an item metadata in a Directory.Build.targets file:

```xml
<ItemGroup>
  <KnownFrameworkReference Update="@(KnownFrameworkReference->HasMetadata('TargetingPackVersion')->HasMetadata('DefaultRuntimeFrameworkVersion'))"
                           TargetingPackVersion="%(KnownFrameworkReference.DefaultRuntimeFrameworkVersion)" />
</ItemGroup>
```

That logic applies to all framework reference items. If we only want to use a X.0.0 targeting pack for a specific set of framework references, then you might want to adapt the above code slightly.

MichaelSimons:
> We should fix that by modifying the KnownFrameworkReference item metadata to point to the package version that source build produces instead of what the SDK decides to target. We do something similar in dotnet/runtime to make sure that the live built asset is chosen.

Thanks @ViktorHofer - that seems like just the right solution we want here.

mthalman:
Ok, I made the change but there's still an issue.

Prior to the change, there were prebuilts for `Microsoft.NETCore.App.Ref.6.0.13` and `Microsoft.AspNetCore.App.Ref.6.0.13`. But now after the change I get a prebuilt for `Microsoft.AspNetCore.App.Ref.3.0.0`. Does that make sense to anyone?

mthalman:
Ah, so we've got this metadata:
image

But in this case we do want the 3.0.1 version. Should I exclude this `KnownFrameworkReference` change for `netcoreapp3.0`?

ViktorHofer:
https://github.com/dotnet/source-build-reference-packages/tree/main/src/targetPacks/ILsrc/microsoft.aspnetcore.app.ref

There is no 3.0.0 targeting pack checked into aspnetcore. It would probably be best to build a lookup table that includes FrameworkReference and TargetingPackVersion for all the checked-in targeting packs.

ViktorHofer:
Opened https://github.com/dotnet/source-build-reference-packages/pull/572 to resolve the issues with implicit framework references once and for all :)

src/referencePackages/src/system.text.json/6.0.0/lib/net6.0/System.Text.Json.cs
```diff
@@ -7,7 +7,7 @@
 [assembly: System.Runtime.CompilerServices.CompilationRelaxations(8)]
 [assembly: System.Runtime.CompilerServices.RuntimeCompatibility(WrapNonExceptionThrows = true)]
 [assembly: System.Diagnostics.Debuggable(System.Diagnostics.DebuggableAttribute.DebuggingModes.IgnoreSymbolStoreSequencePoints)]
-// Commenting out because JsonSerializerOptionsUpdateHandler is an internal type
+// Commenting out because JsonSerializerOptionsUpdateHandler is an internal type (https://github.com/dotnet/sdk/issues/31098)
```
ViktorHofer, line 10: The assembly attribute `[assembly: System.Reflection.Metadata.MetadataUpdateHandler(typeof(JsonSerializerOptionsUpdateHandler))]` shouldn't be generated anymore with the GenAPI used in main. Can you please remove the comments?

mthalman:
Ugh, this is still generating a prebuilt somehow: `Microsoft.AspNetCore.App.Ref.6.0.13`

src/referencePackages/src/system.text.json/6.0.0/lib/net6.0/System.Text.Json.cs
mthalman: No, I mistakenly added it back. Regenerating it removes this content. Updated.

ViktorHofer:
I see that you found a fix to unblock the PR but do we understand why adding these four packages brings in a dependency to Microsoft.AspNetCore.App.Ref/6.0.13?

mthalman:
> I see that you found a fix to unblock the PR but do we understand why adding these four packages brings in a dependency to Microsoft.AspNetCore.App.Ref/6.0.13?

No idea. The binlog is here: https://dev.azure.com/dnceng-public/public/_build/results?buildId=206035&view=results. I'm not seeing how it's getting that.

ViktorHofer:
Found it:

image

Whenever targeting net6.0, apparently the aspnetcore.app.ref targeting pack is added as a dependency. Investigating further.

ViktorHofer:
@mthalman can you please add the following property somewhere to the Directory.Build.targets file: `<DisableTransitiveFrameworkReferenceDownloads>true</DisableTransitiveFrameworkReferenceDownloads>`? That disable the aspnetcore targeting packs transitively being downloaded when they aren't used.

ViktorHofer (approved):
Looks great

MichaelSimons:
@mthalman, FYI I discovered there is an [issue](https://github.com/dotnet/source-build/issues/3328) with the generator regarding the assembly version.  I mention this in case if affects your usage of these new packages.

PR #569, opened 2023-03-10
Remove all netcoreapp3.0 artifacts

MichaelSimons:
This is an experiment to see if anything in the product references this or if they are obsolete and can be cleaned up.


src/referencePackages/src/system.drawing.common/4.7.2/System.Drawing.Common.4.7.2.csproj
```diff
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>netstandard2.0;netcoreapp3.0;net461</TargetFrameworks>
```
ViktorHofer, line 4: Are you sure this is the right thing to do? This means that any .NETCoreApp library referencing this one will pick the netstandard2.0 asset instead of the netcoreapp3.0 one. The netstandard2.0 public API differs from the netcoreapp3.0 one.

MichaelSimons: This was an experiment in which I made some bad assumptions.  We need to create some tooling that will identity which SBRP packages and TFMs are being referenced in source build.  From this information we can property detect what can be cleaned up from SBRP.

PR #132, opened 2020-07-27
Use StrongNameKeyId rather than directly using AssemblyOriginatorKeyFile: fix PublicKeyToken

dagood:
The projects currently use lines like `<AssemblyOriginatorKeyFile>$(KeyFileDir)MSFT.snk</AssemblyOriginatorKeyFile>` to set the key to use for strong-name signing. This seems to conflict with the Arcade configuration method, which looks like this:

```
    <When Condition="'$(StrongNameKeyId)' == 'Microsoft'">
      <PropertyGroup>
        <AssemblyOriginatorKeyFile>$(MSBuildThisFileDirectory)snk/MSFT.snk</AssemblyOriginatorKeyFile>
        <PublicKey>$(MicrosoftPublicKey)</PublicKey>
        <PublicKeyToken>b03f5f7f11d50a3a</PublicKeyToken>
      </PropertyGroup>
    </When>
```

The result is that the dev branch is building assemblies with the wrong PublicKeyToken. This gets caught when trying to build with bootstrap flow using intermediate nupkgs:

1. On compile phase 1, `NuGet.LibraryModel` compiles using `Microsoft` key against prebuilts from NuGet.org that use `Shared` key in their identities. This means the intermediate nupkg contains a set of DLLs where the identity is `Microsoft` with references to `Shared` identities. This already won't work if used: the output doesn't fulfill its own references.

2. On compile phase 2, `NuGet.LibraryModel` compiles using `Microsoft` key against source-built DLLs with `Microsoft` identities and references to `Shared` identities. `Csc` fails with `CS0012` failing to load:

    ```
    Errors
        lib/netstandard2.0/NuGet.Commands.cs(1002,63): error CS0012: The type 'LibraryIdentity' is defined in an assembly that is not referenced. You must add a reference to assembly 'NuGet.LibraryModel, Version=5.1.0.5, Culture=neutral, PublicKeyToken=31bf3856ad364e35'. [/work/artifacts/source-build/self/src/src/referencePackages/src/nuget.commands/5.1.0/NuGet.Commands.5.1.0.csproj]
[...]

PR #573, opened 2023-03-16
Delete snk keys

ViktorHofer:
Follow-up on https://github.com/dotnet/source-build-reference-packages/pull/132 which migrated to use Arcade's strong name key layout.


crummel (approved):

crummel:
Sure, thanks for the fix!

PR #578, opened 2023-03-20
Merge package source generators

ViktorHofer:
Commit 1: Rename ReferencePackageSourceGenerator to PackageSourceGenerator
Commit 2: Enable nullable reference type warnings and simplify GenerateProjects task

Commit 3: Merge text-only and reference package source generators
  - Enables target framework trimming in text-only package nuspecs (same as in reference package sources)
  - Makes both generators use the same logging / restore mechanism (UI improvement)
  - Is a pre-cursor to enable ProjectReferences as text-only packages didn't contain any TFMs
  - Removes duplicated infrastructure

We could discuss if we want to do content file trimming as well in text-only packages based on if those are inside a TFM folder. Doing that would just be for correctness without any meaningful impact.


MichaelSimons (approved):
Thanks for splitting the changes (specifically the renaming) into separate commits.  It made the changes easy to review.  FYI, I tested it out locally on a few text-only packages.

PR #583, opened 2023-03-22
Remove unused targeting pack

MichaelSimons:
Removed microsoft.aspnetcore.app.ref.3.0.1 as it is no longer required by source-build.


oleksandr-didyk (approved):

ViktorHofer (changes requested):
Please also remove this line: https://github.com/dotnet/source-build-reference-packages/blob/df416d414a5dc55cef4e7f5fcc9b5e9d2f855480/src/referencePackages/Directory.Build.targets#L39

MichaelSimons:
> Please also remove this line:
> 
> https://github.com/dotnet/source-build-reference-packages/blob/df416d414a5dc55cef4e7f5fcc9b5e9d2f855480/src/referencePackages/Directory.Build.targets#L39

Done

PR #589, opened 2023-03-28
Fix package source generator nodes lock

ViktorHofer:
The inner builds of the generated multi-targeting source projects that invoke GenAPI collide with each other when different tfms choose the same best compatible package tfm. I.e. `netstandard2.0` and `netcoreapp2.0` in System.Runtime.CompilerServices.Unsafe/4.7.0 both pick the `netstandard2.0` asset as best compatible.

When retrieving the compile items, distinct the item set to avoid duplicates that would then result in inner build nodes locking each other.

Fixes https://github.com/dotnet/source-build/issues/3345

cc @mthalman


oleksandr-didyk (approved):

ViktorHofer:
Before opening this PR I was able to reproduce the issue and captured a binlog in which I saw the duplicate nodes. With the changes, the duplicates inner build nodes don't exist anymore so it should be safe to merge the PR :)

PR #592, opened 2023-03-30
add Microsoft.Build 17.3.2

oleksandr-didyk:
Some of the introduced manual changes:

  - System.Collection.Immutable: 6.0.0 -> https://github.com/dotnet/sdk/issues/31484
  - System.Security.Cryptography.Xml: 6.0.0 -> forced the package to use 6.0.1 of one of its dependencies (was using 6.0.0 originally). Reason: no other ref pack / repo were using 6.0.0, while some were referencing the 6.0.1 one as it contains security fixes
  - System.Security.Permissions: 6.0.0
  		- silenced an API depreciation warning, will create an issue describing it to the GenAPI team
  		- removed generated `net5.0` code since it would not be used by any of the consuming packages
  - System.Security.Cryptography: 6.0.0 -> added missing indexer name attribute, will create and issue describing it to the GenAPI team


eng/Build.props
```diff
@@ -23,8 +23,25 @@
     <DependencyPackageProjects Include="$(RepoRoot)src\referencePackages\src\**\Microsoft.Bcl.AsyncInterfaces.6.0.0.csproj" />
     <DependencyPackageProjects Include="$(RepoRoot)src\referencePackages\src\**\System.Text.Encodings.Web.6.0.0.csproj" />
     <DependencyPackageProjects Include="$(RepoRoot)src\referencePackages\src\**\System.Text.Json.6.0.0.csproj" />
-
+    
+    <DependencyPackageProjects Include="$(RepoRoot)src\referencePackages\src\**\Microsoft.Build.Framework.17.3.2.csproj" />
```
MichaelSimons, line 27: These have to be defined in the correct dependency order in order to prevent prebuilts.  e.g. System.Security.Permissions.6.0.0 must be built before Microsoft.Build.Framework.17.3.2.

oleksandr-didyk: Should be ordered correctly now

eng/Build.props
```diff
@@ -21,10 +21,32 @@
...
+    <DependencyPackageProjects Include="$(RepoRoot)src\referencePackages\src\**\System.Security.Cryptography.ProtectedData.6.0.0.csproj" />
+    <DependencyPackageProjects Include="$(RepoRoot)src\referencePackages\src\**\System.Configuration.ConfigurationManager.6.0.0.csproj" />
+
     <DependencyPackageProjects Include="$(RepoRoot)src\referencePackages\src\**\System.Text.Encodings.Web.6.0.0.csproj" />
     <DependencyPackageProjects Include="$(RepoRoot)src\referencePackages\src\**\System.Text.Json.6.0.0.csproj" />
 
+    <DependencyPackageProjects Include="$(RepoRoot)src\referencePackages\src\**\System.Threading.Tasks.Dataflow.6.0.0.csproj" />
+    <DependencyPackageProjects Include="$(RepoRoot)src\referencePackages\src\**\System.Reflection.MetadataLoadContext.6.0.0.csproj" />
     <DependencyPackageProjects Include="$(RepoRoot)src\referencePackages\src\**\System.Text.Encoding.CodePages.6.0.0.csproj" />
+    <DependencyPackageProjects Include="$(RepoRoot)src\referencePackages\src\**\System.Resources.Extensions.6.0.0.csproj" />
```
MichaelSimons, line 43: Please group all new packages together at the end of the list.  The reason this is important is that when we update previous source-built artifacts, we can remove the entries for the new projects it includes.  New PRs may miss that cut and if they are not grouped at the bottom, it makes this management more difficult..

MichaelSimons: If you don't mind, please update the comment to reflect this convention.

oleksandr-didyk: Of course, will do

MichaelSimons (approved):

oleksandr-didyk:
Added the SBRPs to the repo tracking issue https://github.com/dotnet/source-build/issues/2357#issuecomment-1492301456

PR #593, opened 2023-03-30
remove readme tag from nuspec files

oleksandr-didyk:
When generating an SBRP, the tooling copies the `*.nuspec` file with some minor changes. The file in question can contain a `<readme>` tag pointing to the the `README.md` file.

Since the tooling does not copy the `README.md` file itself, when the project is build an exception is thrown (specifically, [NU5039](https://learn.microsoft.com/en-us/nuget/reference/errors-and-warnings/nu5039)). Example of this exception for `Microsoft.Build: 17.3.2`:
```
error NU5039: The readme file 'README.md' does not exist in the package.
```

To address this exception, this PR add functionality for removing the tag from the `*.nuspec`

### DIFF
--- a/src/packageSourceGenerator/PackageSourceGenerator.proj
+++ b/src/packageSourceGenerator/PackageSourceGenerator.proj
@@ -295,6 +295,7 @@
                    IncludeTargetFrameworks="$(PackageTargetFrameworks)"
                    ExcludeTargetFrameworks="$(ExcludeTargetFrameworks)"
                    RemoveIcon="true"
+                   RemoveReadMe="true"
                    RemoveRuntimeSpecificDependencies="true" />
 
     <Message Text="$(MSBuildProjectName) -> $(NuspecTargetPath)"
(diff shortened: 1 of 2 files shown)


src/packageSourceGenerator/PackageSourceGenerator.proj
```diff
@@ -295,6 +295,7 @@
                    IncludeTargetFrameworks="$(PackageTargetFrameworks)"
                    ExcludeTargetFrameworks="$(ExcludeTargetFrameworks)"
                    RemoveIcon="true"
+                   RemoveReadMe="true"
```
oleksandr-didyk, line 298: Kept the implementation the same as before, just added one more option
Not sure if we want to simplify this somehow or maybe its not worth it

mthalman:
What is the need to remove the readme tag? What is being fixed by this change?

oleksandr-didyk:
@mthalman That's on me, I forgot to add a proper description to the PR. Added now

mthalman (approved):

MichaelSimons (approved):

PR #596, opened 2023-04-04
Remove all net451 references

MichaelSimons:
The PR contains the following changes:

1. Removed the net451 targeting pack
2. Added an net451 exclusion to the generate tooling so that it will no longer include this tfm.
3. Removed net451 from:
    1. System.Data.Common/4.3.0
    2. System.Data.SqlClient/4.8.5

I wanted to simply regenerated the aforementioned packages that had net451 removed from them but encountered https://github.com/dotnet/sdk/issues/31570.  There were too many manual edits needed.


mthalman (commented):
Need to clean up the nuspec files too. Search for ".NETFramework4.5.1"

MichaelSimons:
> Need to clean up the nuspec files too. Search for ".NETFramework4.5.1"

Done.

PR #609, opened 2023-04-06
Add misc reference packages to eliminate prebuilts

MichaelSimons:
These are the SBRP packages being referenced by the various repos in main as reported in the prebuilt report.


mmitche (approved):

mthalman:
Looks like this fixes https://github.com/dotnet/source-build/issues/3348

mthalman (approved):

PR #630, opened 2023-04-12
Remove reserved "-e" bash shortcut

ViktorHofer:
Closes https://github.com/dotnet/source-build/issues/3360

The "-e" option is a reserved argument and can't be used. I decided to just remove it and add some more error validation and logging for better UX. To keep both arguments in sync, I also decided to remove it from the Windows counterpart.

### DIFF
--- a/eng/generate.ps1
+++ b/eng/generate.ps1
@@ -32,7 +32,7 @@ function Get-Help() {
   Write-Host "                                               option above, one per line. If specified, the -package option is ignored."
   Write-Host "  -d|-destination                              A path to the root of the repo to copy source into."
   Write-Host "  -t|-type                                     Type of the package to generate. Accepted values: ref (default) | text."
-  Write-Host "  -x|-excludeDependencies                         Determines if package dependencies should be excluded. Default is false."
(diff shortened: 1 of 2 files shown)


generate.sh
mthalman, line 1: Should remove the `-e` option from the help text.

mthalman (approved):

eng/generate.ps1
```diff
@@ -32,7 +32,7 @@ function Get-Help() {
   Write-Host "                                               option above, one per line. If specified, the -package option is ignored."
   Write-Host "  -d|-destination                              A path to the root of the repo to copy source into."
   Write-Host "  -t|-type                                     Type of the package to generate. Accepted values: ref (default) | text."
-  Write-Host "  -e|-excludeDependencies                      Determines if package dependencies should be excluded. Default is false."
+  Write-Host "  -excludeDependencies                         Determines if package dependencies should be excluded. Default is false."
```
MichaelSimons, line 35: I personally use the aliases in my local dev workflows.  I utilize this option frequently.  Consider changing the alias to a different alias such as 'x'.

ViktorHofer: Updated and tested locally. "-x" works as expected.

MichaelSimons (approved):

PR #631, opened 2023-04-12
Ignore external packages in package source generation

ViktorHofer:
Fixes https://github.com/dotnet/source-build/issues/3364

### DIFF
--- a/src/packageSourceGenerator/PackageSourceGenerator.proj
+++ b/src/packageSourceGenerator/PackageSourceGenerator.proj
@@ -328,10 +328,16 @@
       <TargetingPackage Include="Microsoft.AspNetCore.App.Ref" />
       <TargetingPackage Include="Microsoft.WindowsDesktop.App.Ref" />
     </ItemGroup>
+    
+    <!-- External packages aren't visited as they shouldn't be added to SBRP. --> 
+    <ItemGroup>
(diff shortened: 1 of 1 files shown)


MichaelSimons (approved):

src/packageSourceGenerator/PackageSourceGenerator.proj
```diff
@@ -328,10 +328,16 @@
       <TargetingPackage Include="Microsoft.AspNetCore.App.Ref" />
       <TargetingPackage Include="Microsoft.WindowsDesktop.App.Ref" />
     </ItemGroup>
+    
+    <!-- External packages aren't visited as they shouldn't be added to SBRP. --> 
+    <ItemGroup>
+      <ExternalPackage Include="Newtonsoft.Json" />
```
mthalman, line 334: @MichaelSimons - Does this need to be version-specific at all? If/when there's a 14.0.0 version of Newtonsoft.Json, wouldn't we want to support repos continuing to reference the older 13.0.1 version?

MichaelSimons: I don't think so.  To date we have never needed Newtonsoft.Json in SBRP.  If multiple versions without CVEs are supported and required then we would add them to source-build-externals.

oleksandr-didyk (approved):

PR #633, opened 2023-04-13
Regenerate System.Data.SqlClient for net6.0 only

ViktorHofer:
SBRP wants to remove targeting packs for older TFMs like netcoreapp2.1. Upgrade the netcoreapp2.1 TFM to net6.0. dotnet/runtime is the single consumer and I will submit a PR to runtime to pick the right TFM under source-build.

While re-generating the source, found a GenAPI bug and filed https://github.com/dotnet/sdk/issues/31774 for it.


src/referencePackages/src/system.data.sqlclient/4.8.5/ref/net6.0/System.Data.SqlClient.cs
```diff
@@ -4,354 +4,630 @@
...
+
         public string SourceColumn { get { throw null; } set { } }
+
         public int SourceOrdinal { get { throw null; } set { } }
     }
-    public sealed partial class SqlBulkCopyColumnMappingCollection : System.Collections.CollectionBase
+
+    public sealed partial class SqlBulkCopyColumnMappingCollection : Collections.CollectionBase
     {
-        internal SqlBulkCopyColumnMappingCollection() { }
```
ViktorHofer, line 339: @andriipatsula it looks like GenAPI drops the internal ctor: https://github.com/dotnet/corefx/blob/b86ce69c316b4b088356b89bab80f7fa79ff799c/src/System.Data.SqlClient/ref/System.Data.SqlClient.cs#L313

Without the internal ctor, an object can be instantiated. Filed https://github.com/dotnet/sdk/issues/31776

MichaelSimons (approved):

MichaelSimons:
@ViktorHofer - We will have to coordinate merging this PR with the corresponding runtime change otherwise this SBRP dependency flow into installer will be blocked.  Given the high number of SBRPs being added right now, I would like to avoid a long delay in SBRP flows into installer.

ViktorHofer:
Sure. Here's the change: https://github.com/dotnet/runtime/pull/84782

MichaelSimons (approved):

PR #662, opened 2023-05-04
Add DiaSymReader 2.0.0 reference package

333fred:


MichaelSimons:
This has been addressed in https://github.com/dotnet/sdk/pull/32227.  We do not want to add a reference package for something we are already source-building.

PR #673, opened 2023-05-16
remove nuget 6.2.2

oleksandr-didyk:
Required by https://github.com/dotnet/runtime/pull/86151#issue-1707704531 and https://github.com/dotnet/roslyn/pull/67436#issuecomment-1549946447

The SBRP version of Nuget gets included in packages produced by Arcade. When some of these packages are then used by repo's like `runtime`, they try to load in the packaged Nuget, resulting in an exception (see https://github.com/dotnet/runtime/pull/86151#issuecomment-1549669873). 

This PR removes the affected version with Arcade having it baselined until an intermediate for latest version will be available (see https://github.com/dotnet/arcade/pull/13027/files#diff-f88ce4026366ce74fb1a981785c542757617abf4393a33ca34465a5e9a0dd343R5)


MichaelSimons (approved):

ViktorHofer:
Presumably, tasks packages don't need to bundle NuGet assemblies at all. NuGet is already part of the SDK and available when invoked via Desktop's msbuild. @rainersigwald @dsplaisted @nkolev92 can you please double check my assumption?

rainersigwald:
For `msbuild.exe`/VS I don't think we can automatically find NuGet on behalf of a task today, but `dotnet msbuild` would get it for free.

ViktorHofer:
That should be good enough for source build. I will update the Arcade packages and ApiCompat/GenAPI in dotnet/sdk to not bundle NuGet assemblies in the package for .NETCoreApp.

mthalman:
The removal of these 6.2.2 packages breaks the per-repo-PVP work for Arcade. Arcade is one of the first things that gets built in SB. So if it can't use SBRP the only other option is PSB (6.7.0-preview... instead of 6.2.2). Would such a newer version be acceptable for repos consuming Arcade?

ViktorHofer:
We don't bundle NuGet assemblies into the Arcade packages for .NETCoreApp anymore. Are you sure that they are still required for sourcebuild?

mthalman:
> We don't bundle NuGet assemblies into the Arcade packages for .NETCoreApp anymore. Are you sure that they are still required for sourcebuild?

@ViktorHofer -
There's a compile-time dependency on this version of the NuGet packages: https://github.com/dotnet/arcade/blob/e375d9151b2345e26a0bf967913fa4d621891496/src/Microsoft.DotNet.PackageTesting/Microsoft.DotNet.PackageTesting.csproj#L12

Are you saying it should be safe to add back NuGet 6.2.2 to SBRP and things won't be broken now because Arcade won't pass on ref assemblies in its packages anymore?

ViktorHofer:
Exactly. Arcade now only needs NuGet assemblies to compile against the contract, without invoking or bundling them.

PR #676, opened 2023-05-16
Remove unreferenced targeting packs

MichaelSimons:
Removed the following targeting packs:

1. microsoft.aspnetcore.app.ref/6.0.0
2. microsoft.aspnetcore.app.ref/7.0.0
3. netstandard.library/2.0.0


MichaelSimons:
> Please remove the corresponding entries under referencePackages as well: https://github.com/dotnet/source-build-reference-packages/blob/main/src/referencePackages/Directory.Build.targets#L41-L42

Done in https://github.com/dotnet/source-build-reference-packages/pull/677

PR #689, opened 2023-06-01
Add SBRPTests

v-chayan:
Implement some basic tests. Resolves https://github.com/dotnet/source-build/issues/3089


run-tests.sh
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env bash
```
MichaelSimons, line 1: IMO this should be merged into the build script.  It should be exposed with a `--test` option.  This give the build experience a more common feel/

tests/SBRPTests/SBRPTests.cs
```diff
@@ -0,0 +1,57 @@
+using System;
+using System.Runtime.InteropServices;
+using Xunit;
+
+namespace SBRPTests;
+
+public class ReferencePackagesTest : IClassFixture<SBRPFixture>, IDisposable
```
MichaelSimons, line 7: This name doesn't accurately reflect what the test class does.  It is not validating the reference packages themselves rather it is testing the generator.  Additionally it looks like the current test is validating both reference and text only packages.

tests/SBRPTests/PackageData.cs
```diff
@@ -0,0 +1,20 @@
+using System.Collections.Generic;
+
+namespace SBRPTests;
+
+public class PackageData
```
MichaelSimons, line 5: Do you see this being used across multiple tests?  If not consider utilizing MemberData via a class instance within the consuming test case.  Getting the test data closer to the place where it is used helps improve the UX/maintainability.

tests/SBRPTests/SBRPFixture.cs
```diff
@@ -0,0 +1,16 @@
...
+
+public class SBRPFixture
+{
+    public string WorkingDirectory { get; set; }
+
+    public SBRPFixture()
+    {
+        WorkingDirectory = Environment.CurrentDirectory = Environment.CurrentDirectory.Substring(0, Environment.CurrentDirectory.IndexOf("source-build-reference-packages") + 31);
+        CommandHelper.CommandOutput("git", "reset HEAD", WorkingDirectory);
+        CommandHelper.CommandOutput("git", "checkout src/", WorkingDirectory);
```
MichaelSimons, line 13: Won't this lose uncommitted changes on disk?  My initial thought was the tests should utilize a sandboxed location for the generated output and compare it against the actual content.  It is hypothetically possible that we could overwrite uncommitted user customization to the package we are testing against.

tests/SBRPTests/SBRPTests.cs
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    [Theory(Timeout = 300000)]    
+    [MemberData(nameof(PackageData.Data), MemberType = typeof(PackageData))]
+    public void VerifySBRP(string package, PackageType type)
+    {
+        switch (type)
+        {
+            case PackageType.Reference:
+                if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
```
MichaelSimons, line 23: There is duplication between the window/linux as well as reference/text paths.  For instance the invocation of generatre.cm/sh as well as the validation logic.  To increase maintainability of this code, please refactor to eliminate this duplication.

tests/SBRPTests/SBRPFixture.cs
```diff
@@ -0,0 +1,16 @@
...
+
+namespace SBRPTests;
+
+public class SBRPFixture
+{
+    public string WorkingDirectory { get; set; }
+
+    public SBRPFixture()
+    {
+        WorkingDirectory = Environment.CurrentDirectory = Environment.CurrentDirectory.Substring(0, Environment.CurrentDirectory.IndexOf("source-build-reference-packages") + 31);
```
MichaelSimons, line 11: I prefer avoiding side effects from running the tests - e.g. Environment.CurrentDirectory.  This doesn't feel necessary. This is necessary if you ever want to run the tests in synchronously which is useful for performance reasons.

tests/SBRPTests/SBRPTests.cs
```diff
@@ -0,0 +1,57 @@
...
+
+                CommandHelper.CommandOutput("git", "add .", WorkingDirectory);
+                Assert.DoesNotContain("modified:   src/", CommandHelper.CommandOutput("git", "status", WorkingDirectory));
+                break;
+        }
+    }
+
+    public void Dispose()
+    {
+        CommandHelper.CommandOutput("git", "reset HEAD", WorkingDirectory);
```
MichaelSimons, line 53: Please try to avoid duplication.  This logic exists in the test fixture

tests/SBRPTests/SBRPTests.cs
```diff
@@ -0,0 +1,57 @@
...
+
+    public string WorkingDirectory { get; set; }
+    public ReferencePackagesTest(SBRPFixture fixture)
+    {
+        WorkingDirectory = fixture.WorkingDirectory;
+    }
+
+    [Theory(Timeout = 300000)]    
+    [MemberData(nameof(PackageData.Data), MemberType = typeof(PackageData))]
+    public void VerifySBRP(string package, PackageType type)
```
MichaelSimons, line 18: Consider a more descriptive name - VerifyGenerateScript

tests/SBRPTests/SBRPTests.cs
```diff
@@ -0,0 +1,57 @@
...
+public class ReferencePackagesTest : IClassFixture<SBRPFixture>, IDisposable
+{
+
+    public string WorkingDirectory { get; set; }
+    public ReferencePackagesTest(SBRPFixture fixture)
+    {
+        WorkingDirectory = fixture.WorkingDirectory;
+    }
+
+    [Theory(Timeout = 300000)]
```
MichaelSimons, line 16: This seems like a really long timeout.  What is the default?  Have you encountered this being an issue?

tests/SBRPTests/PackageData.cs
```diff
@@ -0,0 +1,20 @@
...
+{
+    public static IEnumerable<object[]> Data => new List<object[]>
+    {
+        new object[] { "Microsoft.CodeAnalysis.CSharp.Workspaces,4.5.0", PackageType.Reference },
```
MichaelSimons, line 9: What are the reasons for running these 3 ref packages?  If there are specific scenarios it would be useful to call them out.  If there aren't I would be curious why we 3 different packages are run versus 1.

tests/SBRPTests/CommandHelper.cs
```diff
@@ -0,0 +1,35 @@
+using System;
+using System.Diagnostics;
+using System.Runtime.InteropServices;
+
+namespace SBRPTests;
+internal class CommandHelper
```
MichaelSimons, line 6: Consider coping https://github.com/dotnet/installer/blob/main/src/SourceBuild/content/test/Microsoft.DotNet.SourceBuild.SmokeTests/ExecuteHelper.cs.  One nice thing about it is that is logs what commands are being executed as well as their output.  This is useful when a test fails in that you can look at the output to help diagnose the issue without having to reproduce the issue locally.

tests/SBRPTests/SBRPTests.cs
```diff
@@ -0,0 +1,57 @@
...
+                {
+                    CommandHelper.CommandOutput("generate.cmd", string.Format("-type text -package {0} -excludedependencies", package), WorkingDirectory);
+                }
+                else
+                {
+                    CommandHelper.CommandOutput("generate.sh", string.Format("--type text --package {0} --excludedependencies", package), WorkingDirectory);
+                }
+
+                CommandHelper.CommandOutput("git", "add .", WorkingDirectory);
+                Assert.DoesNotContain("modified:   src/", CommandHelper.CommandOutput("git", "status", WorkingDirectory));
```
MichaelSimons, line 46: Ideally the test failure output would give details information about what the differences were.  Emitting the diff output would be ideal.

MichaelSimons (commented):
Thanks for the PR that introduces tests.  This will be helpful in preventing regressions.

Friendly reminder to link PRs to the issues they are addressing.

This new test should be integrated into CI.  Builds/PR validation should fail if the tests don't pass.

run-tests.sh
v-chayan: Nikola said you were disccussed and the plan is set a new property 'RunTests' which is similar to generate.sh. The test project should only be added if new property 'RunTests' is set. That's why we need this new script.

tests/SBRPTests/CommandHelper.cs
v-chayan: Got it. Will update with this.

tests/SBRPTests/PackageData.cs
v-chayan: These packages are just examples which I picked ramdomly. We could discuss which packages should be covered. Maybe according to the priority of each package.

tests/SBRPTests/SBRPTests.cs
v-chayan: Timout feature seems not work. I tried to set it to 1 and the tests still could finish correctly. So just removed the timeout.

tests/SBRPTests/SBRPFixture.cs
v-chayan: The original request I got is "Create some light-weight, basic, tests, using XUnit. For the start, we’d want to have few basic tests that just generate existing reference packages and use git to see if there are any open files". So I need to do this cleanup between different tests. I'll discuss with @NikolaMilosavljevic to see if we need to change current implementation

tests/SBRPTests/SBRPTests.cs
v-chayan: I will update it to GenerateScriptRegressionTests

tests/SBRPTests/SBRPFixture.cs
MichaelSimons: My primary concern is that the test can't revert changes the user may have locally.

tests/SBRPTests/SBRPTests.csproj
```diff
@@ -0,0 +1,16 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>$(NetCurrent)</TargetFramework>
+    <Nullable>enable</Nullable>
+    <IsPackable>false</IsPackable>
```
ViktorHofer, line 6: That's already the default in arcadified repositories.

```suggestion
```

tests/SBRPTests/SBRPTests.csproj
```diff
@@ -0,0 +1,16 @@
...
+    <IsUnitTestProject>true</IsUnitTestProject>
```
ViktorHofer, line 7: Arcade already offers an in-built way to define test projects: https://github.com/dotnet/arcade/blob/234e0726c7384ee84bf08550f2d16a1ff2d5c543/src/Microsoft.DotNet.Arcade.Sdk/tools/Tests.props#L16

Please rename this project to something `.Tests.csproj` so that Arcade automatically recognizes it. Also remove the package references below, including the Versions.props entries for it. All of that is already handled by arcade.

ViktorHofer (changes requested):
Please change the test project to use Arcade's defaults to avoid all the boilerplate code. Also, consider using the default build scripts for running tests and test their behavior not just for Unix but also for Windows.

tests/SBRPTests/SBRPFixture.cs
NikolaMilosavljevic: This would be easily achievable using `-d` option of the generation script.

However, if user has uncommitted customization changes in the exact same project, the test will fail. Tests might also be invalid for any project that was modified after original generation.

An alternative approach might be to check for any uncommitted changes and either 1) not run tests until those are committed, or 2) stash and later reapply the changes.

run-tests.sh
NikolaMilosavljevic: We did discuss this offline and while the original plan (in PR) would have worked, I see the benefit of fitting 'test' experience within the `build.sh` script. It would require some tweaks in `eng/Build.props`.

tests/SBRPTests/SBRPFixture.cs
NikolaMilosavljevic: @v-chayan running tests in sandbox is preferable. We should not revert any user's changes. We could have a warning that there are uncommitted changes in the projects that we're regenerating.

v-chayan: Got it. Will update the implementation

tests/SBRPTests/SBRPTests.csproj
v-chayan: Got it. Thanks for letting me know this feature of Arcade!

tests/SBRPTests/SBRPFixture.cs
v-chayan: Updated. Now it won't change Environment.CurrentDirectory.

eng/Versions.props
```diff
@@ -21,5 +21,10 @@
     <MicrosoftNETCoreILDAsmVersion>6.0.0-preview.6.21352.12</MicrosoftNETCoreILDAsmVersion>
     <!-- SDK dependencies -->
     <MicrosoftDotNetGenAPITaskPackageVersion>8.0.100-preview.5.23273.9</MicrosoftDotNetGenAPITaskPackageVersion>
+    <!-- Test SDK dependencies -->
+    <MicrosoftNETTestSdkVersion>16.11.0</MicrosoftNETTestSdkVersion>
+    <!-- Xunit dependencies -->
+    <XunitVersion>2.4.1</XunitVersion>
+    <XunitRunnerConsoleVersion>2.4.1</XunitRunnerConsoleVersion>
```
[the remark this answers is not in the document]
v-chayan: Updated

run-tests.sh
v-chayan: Updated. Now just use build.sh --test to launch the test

tests/GenerateScriptRegressionTests/GenerateScriptRegressionTests.Tests.csproj
```diff
@@ -0,0 +1,8 @@
+<Project Sdk="Microsoft.NET.Sdk">
```
ViktorHofer, line 1: nit: We usually don't use "regression test" to name test directories and projects. IMO renaming the folder and the project to `GenerateScriptTests` and `GenerateScript.Tests.csproj` would be better.

tests/GenerateScriptRegressionTests/GenerateScriptRegressionTestsFixture.cs
```diff
@@ -0,0 +1,14 @@
...
+
+namespace GenerateScriptRegressionTests;
+
+public class GenerateScriptRegressionTestsFixture
+{
+    public string WorkingDirectory { get; set; }
+
+    public GenerateScriptRegressionTestsFixture()
+    {
+        WorkingDirectory = Environment.CurrentDirectory.Substring(0, Environment.CurrentDirectory.IndexOf("source-build-reference-packages") + 31);
```
MichaelSimons, line 12: I find the `WorkingDirectory` terminology here confusing.  To me the working directory for the `VerifyGenerateScript`  is the sandbox folder.  For other tests in the future it may be the `bin` directory.  To me this is really the `SBRPRoot` or `RepoRoot` directory.

tests/GenerateScriptRegressionTests/GenerateScriptRegressionTestsFixture.cs
```diff
@@ -0,0 +1,14 @@
+﻿using System;
+using Xunit.Abstractions;
+
+namespace GenerateScriptRegressionTests;
+
+public class GenerateScriptRegressionTestsFixture
```
MichaelSimons, line 6: Utilizing a TestFixture for static Environmental information feels like overkill.  There is no need for the fixture overhead for static information.

tests/GenerateScriptRegressionTests/GenerateScriptRegressionTests.cs
```diff
@@ -0,0 +1,65 @@
...
+
+public class GenerateScriptRegressionTests : IClassFixture<GenerateScriptRegressionTestsFixture>, IDisposable
+{
+    private string sandbox_folder = Path.Combine(Environment.CurrentDirectory,"sandbox_folder");
+    public ITestOutputHelper output { get; set; }
+    public string WorkingDirectory { get; set; }
+    public GenerateScriptRegressionTests(GenerateScriptRegressionTestsFixture fixture, ITestOutputHelper output)
+    {
+        WorkingDirectory = fixture.WorkingDirectory;        
+        ExecuteHelper.ExecuteProcess("mkdir", "sandbox_folder", output);
```
MichaelSimons, line 35: To improve maintainability, the local `sandbox_folder` variable should be utilized here.

tests/GenerateScriptRegressionTests/GenerateScriptRegressionTests.cs
```diff
@@ -0,0 +1,65 @@
...
+
+public enum PackageType
+{
+    Reference,
+    Text
+}
+
+public class GenerateScriptRegressionTests : IClassFixture<GenerateScriptRegressionTestsFixture>, IDisposable
+{
+    private string sandbox_folder = Path.Combine(Environment.CurrentDirectory,"sandbox_folder");
```
MichaelSimons, line 29: Inconsistent naming convention here here - should be sandboxFolder.

tests/GenerateScriptRegressionTests/GenerateScriptRegressionTests.cs
MichaelSimons, line 35: Instead of calling out of proc, consider utilizing the .NET Framework API to create the folder.

tests/GenerateScriptRegressionTests/GenerateScriptRegressionTests.cs
MichaelSimons, line 29: To be more consistent with the API, I would suggest utilizing `Directory` instead of `Folder`.

tests/GenerateScriptRegressionTests/GenerateScriptRegressionTests.cs
```diff
@@ -0,0 +1,65 @@
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Runtime.InteropServices;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace GenerateScriptRegressionTests;
+
+public class PackageData
```
MichaelSimons, line 10: There is no need to separate this into a separate class.  I would only do that if the data is utilizing across multiple test classes.

tests/GenerateScriptRegressionTests/CompareDirs.cs
```diff
@@ -0,0 +1,86 @@
...
+        while ((file1byte == file2byte) && (file1byte != -1));
+
+        fs1.Close();
+        fs2.Close();
+
+        return ((file1byte - file2byte) == 0);
+    }
+}
+
+class FileCompare : IEqualityComparer<FileInfo>
```
MichaelSimons, line 65: All types should have an access modifier.

MichaelSimons, line 65: My personal preference is one type per file except for nested types.  It improves the readability/discoverability of the code base.

tests/GenerateScriptRegressionTests/CompareDirs.cs
```diff
@@ -0,0 +1,86 @@
...
+    public static bool Compare(string pathA, string pathB, ITestOutputHelper outputHelper)
+    {
+
+        DirectoryInfo dir1 = new DirectoryInfo(pathA);
+        DirectoryInfo dir2 = new DirectoryInfo(pathB);
+
+        List<FileInfo> list1 = dir1.GetFiles("*.*", SearchOption.AllDirectories).ToList();
+        List<FileInfo> list2 = dir2.GetFiles("*.*", SearchOption.AllDirectories).ToList();
+
+        FileCompare myFileCompare = new FileCompare();
```
MichaelSimons, line 19: The use of multiple things called FileCompare (both a class and a method) in this method are confusing to me.

tests/GenerateScriptRegressionTests/CompareDirs.cs
```diff
@@ -0,0 +1,86 @@
...
+            for(int i = 0; i<list1.Count();i++)
+            {
+                areIdentical &= FileCompare(list1[i].FullName, list2[i].FullName);
+                outputHelper.WriteLine(string.Format("File {0} is modified!",list1[i].FullName));
+            }
+        }
+        return areIdentical;
+    }
+
+    private static bool FileCompare(string file1, string file2)
```
MichaelSimons, line 34: In the primary source build tests we have utilize git to diff files as it provides a better UX for indicating the differences.  Consider utilizing it here - ExecuteHelper.ExecuteProcess("git", $"diff --no-index {file1Path} {file2Path}", outputHelper);

MichaelSimons (commented):
I like the utilizing of the arcade testing infra versus adding a new run test script.

tests/GenerateScriptRegressionTests/CompareDirs.cs
v-chayan: Changed the class name to FileInfoComparer

tests/GenerateScriptRegressionTests/CompareDirs.cs
v-chayan: Changed both class to have same access modifier.

tests/GenerateScriptRegressionTests/GenerateScriptRegressionTests.Tests.csproj
v-chayan: Updated.

tests/GenerateScriptRegressionTests/GenerateScriptRegressionTests.cs
v-chayan: Have put them into test class

tests/GenerateScriptRegressionTests/GenerateScriptRegressionTests.cs
v-chayan: Changed to sandboxDirecotry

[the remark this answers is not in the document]
v-chayan: Updated.

tests/GenerateScriptRegressionTests/GenerateScriptRegressionTests.cs
v-chayan: Updated.

v-chayan: Updated.

tests/GenerateScriptRegressionTests/GenerateScriptRegressionTestsFixture.cs
v-chayan: TestFixture was used for cleanup environment because in the first design, we just use git command to check file changes. But now since we're using sandbox design, we could get rid of this TestFixture. Will update.

tests/GenerateScriptRegressionTests/GenerateScriptRegressionTestsFixture.cs
v-chayan: Changed to RepoRoot

tests/GenerateScriptRegressionTests/CompareDirs.cs
v-chayan: I tried the git diff command. But will got many noise such as "warning: CRLF will be replaced by LF ..." and "The file will have its original line endings in your working directory".
I also tried the switch --ignore-cr-at-eol but it seems not work.

tests/GenerateScriptTests/GenerateScriptTests.Tests.csproj
```diff
@@ -0,0 +1,8 @@
+<Project Sdk="Microsoft.NET.Sdk">
```
ViktorHofer, line 1: Infrastructure wise looks good aside from one last nit from my side, please also rename the project file to `GenerateScript.Tests.csproj`.

tests/GenerateScriptTests/CompareDirs.cs
```diff
@@ -0,0 +1,122 @@
...
+using System.Linq;
+using Xunit.Abstractions;
+
+namespace GenerateScriptTests;
+
+public class CompareDirs
+{
+    public static bool Compare(string pathA, string pathB, ITestOutputHelper outputHelper)
+    {
+
```
ViktorHofer, line 15:
```suggestion
```

tests/GenerateScriptTests/CompareDirs.cs
```diff
@@ -0,0 +1,122 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using Xunit.Abstractions;
+
+namespace GenerateScriptTests;
+
+public class CompareDirs
```
ViktorHofer, line 11: The class shouldn't be public as it doesn't need to be accessible from the outside.

```suggestion
internal class CompareDirs
```

tests/GenerateScriptTests/CompareDirs.cs
```diff
@@ -0,0 +1,122 @@
...
+        while ((file1byte == file2byte) && (file1byte != -1));
+
+        fs1.Close();
+        fs2.Close();
+
+        return ((file1byte - file2byte) == 0);
+    }
+}
+
+public class FileInfoComparer : IEqualityComparer<FileInfo>
```
ViktorHofer, line 101:
```suggestion
internal class FileInfoComparer : IEqualityComparer<FileInfo>
```

tests/GenerateScriptTests/GenerateScriptTests.cs
```diff
@@ -0,0 +1,62 @@
...
+        SandboxDirectory = Path.Combine(Environment.CurrentDirectory, $"GenerateTests-{DateTimeOffset.UtcNow.ToUnixTimeSeconds().ToString()}");
+        Directory.CreateDirectory(SandboxDirectory);
+    }
+
+    [Theory]    
+    [MemberData(nameof(GenerateScriptTests.Data), MemberType = typeof(GenerateScriptTests))]
+    public void VerifyGenerateScript(string package, string version, PackageType type)
+    {
+        string command = Path.Combine(RepoRoot, RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ? "generate.cmd" : "generate.sh");
+        string arguments = string.Format("-p {0} -x -d {1}",string.Format("{0},{1}", package, version), SandboxDirectory);
```
MichaelSimons, line 45: String interpolation would simplify this

```suggestion
        string arguments = $"-p {package},{version{ -x -d {SandboxDirectory}";
```

tests/GenerateScriptTests/GenerateScriptTests.cs
```diff
@@ -0,0 +1,62 @@
...
+        {
+            case PackageType.Reference:
+                packageSrcDirectory = Path.Combine(packageSrcDirectory, "referencePackages", "src");
+                break;
+            case PackageType.Text:
+                arguments += " -t text";
+                packageSrcDirectory = Path.Combine(packageSrcDirectory, "textOnlyPackages", "src");
+                break;
+        }
+        packageSrcDirectory = Path.Combine(packageSrcDirectory, package.ToLower(), version);
```
MichaelSimons, line 58: You could consolidate the logic that builds the packageSrcDirectory 

```suggestion
        packageSrcDirectory = Path.Combine(RepoRoot, "src", package.ToLower(), version);
```

tests/GenerateScriptRegressionTests/GenerateScriptRegressionTests.cs
```diff
@@ -0,0 +1,65 @@
...
+                packageSrcFolder = Path.Combine(packageSrcFolder, "referencePackages", "src");
+                break;
+            case PackageType.Text:
+                arguments += " -t text";
+                packageSrcFolder = Path.Combine(packageSrcFolder, "textOnlyPackages", "src");
+                break;
+        }
+        packageSrcFolder = Path.Combine(packageSrcFolder, package.ToLower(), version);
+        ExecuteHelper.ExecuteProcess(command, arguments, output);
+        Assert.True(CompareDirs.Compare(packageSrcFolder, sandbox_folder, output), string.Format("{0},{1} failed in the test!", package, version));
```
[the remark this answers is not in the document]
MichaelSimons: I would not include the System.Threading.Tasks.Extensions,4.5.4 package in the test.

IMO we need more detailed information emitted that indicates the differences or how to discover the differences.  This is the motivation behind my other comment regarding the use of `git diff`.

tests/GenerateScriptTests/GenerateScriptTests.cs
```diff
@@ -0,0 +1,62 @@
...
+
+    [Theory]    
+    [MemberData(nameof(GenerateScriptTests.Data), MemberType = typeof(GenerateScriptTests))]
+    public void VerifyGenerateScript(string package, string version, PackageType type)
+    {
+        string command = Path.Combine(RepoRoot, RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ? "generate.cmd" : "generate.sh");
+        string arguments = string.Format("-p {0} -x -d {1}",string.Format("{0},{1}", package, version), SandboxDirectory);
+        string packageSrcDirectory = Path.Combine(RepoRoot, "src");
+        string sandboxPackageGeneratedDirecotry = Path.Combine(SandboxDirectory, package.ToLower(), version);
+        switch (type)
```
MichaelSimons, line 48:
```suggestion

        switch (type)
```

tests/GenerateScriptTests/GenerateScriptTests.cs
MichaelSimons, line 58:
```suggestion

        packageSrcDirectory = Path.Combine(packageSrcDirectory, package.ToLower(), version);
```

MichaelSimons (commented):
Friendly reminder to integrate the tests into the pipeline so that they run in PR validiation.

tests/GenerateScriptRegressionTests/GenerateScriptRegressionTests.Tests.csproj
v-chayan: Could you please refresh and try again? 
https://github.com/dotnet/source-build-reference-packages/pull/689/files

tests/GenerateScriptRegressionTests/CompareDirs.cs
MichaelSimons: I opened https://github.com/dotnet/source-build/issues/3500 to address this.

v-chayan: Got it. Then I could wait until the issue fixed and switch to "git diff" for comparison.

tests/GenerateScriptRegressionTests/GenerateScriptRegressionTests.cs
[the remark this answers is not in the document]
v-chayan: Will use git diff for the comparison after endline issue fixed. In the meanwhile, could you please help to pick some packages which are more impotant and should be covered in your opinion?

tests/GenerateScriptRegressionTests/CompareDirs.cs
MichaelSimons: FYI https://github.com/dotnet/sdk/issues/33289 was resolved to you should be able to switch to git diff now.

test.cmd
```diff
@@ -0,0 +1,2 @@
+@echo off
+powershell -ExecutionPolicy ByPass -NoProfile -command "& """%~dp0eng\common\Build.ps1""" -restore -build -test %*"
```
ViktorHofer, line 2: Overall the PR looks really good now. My remaining concern is around the expectations of this script. I would expect it to restore and build the repository and then invoke all the tests (currently only GenerateScriptTests.csproj). But in-fact this only restore, builds and invokes the test project. I think that's OK for now (given that the repository is currently not buildable on Windows) but I wanted to call that out. @MichaelSimons any concerns with the current design?

ViktorHofer (approved):

tests/GenerateScriptTests/GenerateScriptTests.cs
```diff
@@ -0,0 +1,64 @@
...
+        Reference,
+        Text
+    }
+
+    public static IEnumerable<object[]> Data => new List<object[]>
+    {
+        new object[] { "System.Xml.ReaderWriter", "4.0.11", PackageType.Reference },
+        new object[] { "Microsoft.Extensions.Logging.Abstractions", "7.0.1", PackageType.Reference },
+        new object[] { "Microsoft.CodeAnalysis.CSharp", "3.11.0", PackageType.Reference },
+        new object[] { "System.Security.Cryptography.Pkcs", "7.0.2", PackageType.Reference },
```
MichaelSimons, line 26: We should be validating text only packages as well.  All the test infrastructure was defined but no test data for them were added.
```suggestion
        new object[] { "System.Security.Cryptography.Pkcs", "7.0.2", PackageType.Reference },
        new object[] { "Microsoft.Build.NoTargets", "3.7.0", PackageType.Text },
```

test.cmd
MichaelSimons: I am fine with that.

MichaelSimons (approved):

ViktorHofer (changes requested):
You checked in the project file which I think you didn't want to? Also, the file name casing of the project file seems wrong.

MichaelSimons:
> You checked in the project file which I think you didn't want to? Also, the file name casing of the project file seems wrong.

The package was regenerated with the latest tooling which is necessary if we are going to use it as a baseline in the tests.  The casing change is because the tooling uses the package name's casing.

MichaelSimons (changes requested):

ViktorHofer:
> The package was regenerated with the latest tooling which is necessary if we are going to use it as a baseline in the tests. The casing change is because the tooling uses the package name's casing.

I see. So the old checked-in project file had an incorrect casing. Sounds good then.

src/textOnlyPackages/src/microsoft.build.notargets/3.7.0/microsoft.build.notargets.nuspec
```diff
@@ -1,12 +1,11 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
+<?xml version="1.0" encoding="utf-8"?>
 <package xmlns="http://schemas.microsoft.com/packaging/2012/06/nuspec.xsd">
   <metadata>
     <id>Microsoft.Build.NoTargets</id>
     <version>3.7.0</version>
     <authors>Microsoft</authors>
     <license type="expression">MIT</license>
     <licenseUrl>https://licenses.nuget.org/MIT</licenseUrl>
-    <icon>MSBuild-NuGet-Icon.png</icon>
```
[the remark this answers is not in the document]
Winniexu01: I tried deleting the icon file, but the test failed. Because the icon file is still generated by the latest tooling. So I reverted the change and kept the icon file.

MichaelSimons (approved):

PR #697, opened 2023-06-06
add Nuget.Commands and Nuget.Credentials 6.2.2

oleksandr-didyk:
Fixes product pre-builts introduced by https://github.com/dotnet/installer/pull/16592 by adding `Nuget.Commands` and `Nuget.Credentials` `6.2.2`


log.txt
mthalman, line 1: Delete this file

MichaelSimons:
Have you verified Arcade is not loading these?  Given arcade is a toolset repo, there is a good chance they are.

log.txt
oleksandr-didyk: My bad, removed

oleksandr-didyk:
> Have you verified Arcade is not loading these? Given arcade is a toolset repo, there is a good chance they are.

Viktor disabled bundling of `Nuget.*` packages with Arcade's projects as mentioned in https://github.com/dotnet/source-build-reference-packages/pull/673#issuecomment-1569508658

MichaelSimons:
> Viktor disabled bundling of Nuget.* packages with Arcade's projects as mentioned in https://github.com/dotnet/source-build-reference-packages/pull/673#issuecomment-1569508658. That was the original reason we could use SBRPed Nuget for Arcade.

Yes but technically it is possible for a nuget package to reference 6.2.2 which would load the SBRP.

oleksandr-didyk:
> I will test it out with the VMR just to be safe

The build passed, Nuget* pre-builts are not present

MichaelSimons (approved):

PR #718, opened 2023-06-16
Regenerate System.Xml.ReaderWriter 4.0.11

MichaelSimons:
These changes contain some infrastructure changes to support overriding DisableImplicitFrameworkReferences in order to avoid a circular reference with the targeting pack.

### DIFF
--- a/docs/known_generator_issues.md
+++ b/docs/known_generator_issues.md
@@ -7,3 +7,16 @@ error CS8138: Cannot reference 'System.Runtime.CompilerServices.TupleElementName
 ```
 
 Workaround: update the declaration using tuple syntax to define tuple names. Or comment/remove this attribute.
+
+## Circular dependency between reference package and targeting pack
+
+```text
+error NU1108: Cycle detected.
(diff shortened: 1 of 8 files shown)


docs/known_generator_issues.md
```diff
@@ -7,3 +7,16 @@ error CS8138: Cannot reference 'System.Runtime.CompilerServices.TupleElementName
...
+
+```text
+error NU1108: Cycle detected.
+error NU1108:   <Reference project> -> NETStandard.Library x.y.z -> <Reference project>
+```
+
+Workaround: add the following property to the project.
+
+```xml
+<DisableImplicitFrameworkReferences>true</DisableImplicitFrameworkReferences>
```
ViktorHofer, line 21: This setting should only be enabled for netstandard1.x TFMs. It's fine to use StartsWith here as netstandard1.x will never use platforms.

```suggestion
<DisableImplicitFrameworkReferences Condition="$(TargetFramework.StartsWith('netstandard1'))">true</DisableImplicitFrameworkReferences>
```

MichaelSimons:
Closing in favor of https://github.com/dotnet/source-build-reference-packages/pull/719

PR #720, opened 2023-06-16
Add system.security.cryptography.pkcs.7.0.2

JanKrivanek:
### Context

Adding system.security.cryptography.pkcs.7.0.2 as 7.0.0 has known vulns.
Should resolve https://github.com/dotnet/msbuild/pull/8903


JanKrivanek:
Thank you @MichaelSimons!

PR #740, opened 2023-07-06
Add NuGet.*.6.2.4 packages

mthalman:
Updates the package versions due to reporting of CVE-2023-29337 for 6.2.2 version.

There were no packages outside of this set of NuGet packages that had any dependencies on these updated packages, so there was no need to update any references.

Fixes https://github.com/dotnet/source-build/issues/3541

### DIFF
--- a/eng/Build.props
+++ b/eng/Build.props
@@ -89,6 +89,18 @@
     <DependencyPackageProjects Include="$(RepoRoot)src\referencePackages\src\**\NuGet.Credentials.6.6.1.csproj" />
 
     <DependencyPackageProjects Include="$(RepoRoot)src\referencePackages\src\**\System.Security.Cryptography.Pkcs.6.0.4.csproj" />
+
+    <DependencyPackageProjects Include="$(RepoRoot)src\referencePackages\src\**\NuGet.Frameworks.6.2.4.csproj" />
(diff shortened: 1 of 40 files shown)


MichaelSimons:
This is going to have to happen in multiple stages.  

1. Add the new package so that arcade can upgrade
2. Once arcade upgrades and flows into the VMR, remove the vulnerable version.

If done in once stage, SBRP won't be able to flow into the VMR until arcade does.  This would effectively block all other SBRP changes from flowing into the VMR which could be bad and block other work.

mthalman:
I've updated to include these as new packages instead of upgrading 6.2.2 to 6.2.4.

Adds the following packages:
* NuGet.Frameworks.6.2.4
* NuGet.Common.6.2.4
* NuGet.Versioning.6.2.4
* NuGet.Configuration.6.2.4
* NuGet.Packaging.6.2.4
* NuGet.LibraryModel.6.2.4
* NuGet.Protocol.6.2.4
* NuGet.DependencyResolver.Core.6.2.4
* NuGet.ProjectModel.6.2.4
* NuGet.Credentials.6.2.4
* NuGet.Commands.6.2.4

MichaelSimons (approved):

PR #743, opened 2023-07-10
Add NuGet.Frameworks 6.5.0

Evangelink:
Relates to https://github.com/microsoft/vstest/pull/4589


MichaelSimons:
Closing per comments in https://github.com/microsoft/vstest/pull/4589#issuecomment-1629627888.  This is only needed for repo level source-builds therefore it should not be added to SBRP.

PR #745, opened 2023-07-11
Updates for preview 6 release

MichaelSimons:
Update self-reference to the preview 6 version and remove `DependencyPackageProjects` included in preview 6.

This PR can't be merged until the source-build artifacts are updated to preview 6

PR #751, opened 2023-07-21
System.Text.Json 7.0.3

rainersigwald:
Followed the README to add STJ 7.0.3

* `./generate.sh --package System.Text.Json,7.0.3`
* revert modified files
* Fix build errors the same way STJ 7.0.2 did

### DIFF
--- a/eng/Build.props
+++ b/eng/Build.props
@@ -55,6 +55,8 @@
     <DependencyPackageProjects Include="$(RepoRoot)src\referencePackages\src\**\NuGet.ProjectModel.6.2.4.csproj" />
     <DependencyPackageProjects Include="$(RepoRoot)src\referencePackages\src\**\NuGet.Credentials.6.2.4.csproj" />
     <DependencyPackageProjects Include="$(RepoRoot)src\referencePackages\src\**\NuGet.Commands.6.2.4.csproj" />
+
+    <DependencyPackageProjects Include="$(RepoRoot)src\referencePackages\src\**\System.Text.Json.7.0.3.csproj" />
   </ItemGroup>
 
   <ItemGroup Condition="'$(BuildDependencyPackageProjects)' == 'true'">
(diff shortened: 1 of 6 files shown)


Forgind:
I got most of the way through making my own PR before I noticed you'd made this one.

Looks mostly the same as what I would've made, but I also added S.T.J to eng\Build.props. It's in the readme. Did you miss it, or did you intentionally skip it, and if the latter, why?

rainersigwald:
STJ 7.0.2 was not there and it didn't seem to be required (https://dev.azure.com/dnceng-public/public/_build/results?buildId=348238&view=logs&j=2f0d093c-1064-5c86-fc5b-b7b1eca8e66a&t=d2b639a6-cb19-5f1f-66fd-8047f66b3026&l=1861). But I don't really understand why, that definitely tripped me up.

mthalman:
> STJ 7.0.2 was not there and it didn't seem to be required (https://dev.azure.com/dnceng-public/public/_build/results?buildId=348238&view=logs&j=2f0d093c-1064-5c86-fc5b-b7b1eca8e66a&t=d2b639a6-cb19-5f1f-66fd-8047f66b3026&l=1861). But I don't really understand why, that definitely tripped me up.

It does need to be listed there. New packages need to be listed there. Once a .NET release occurs and we can self-reference the latest version of the SBRP package, then we can clean up old references (see https://github.com/dotnet/source-build-reference-packages/pull/745). That's why 7.0.2 isn't listed.

rainersigwald:
Ok. Done.

mthalman (approved):

Forgind:
As far as where in the Build.props, the readme said "dependency order." I'm not aware of any dependencies among the packages in that list, so I don't think its placement matters, and being on bottom is fine. Is that what you were thinking, too?

mthalman:
> As far as where in the Build.props, the readme said "dependency order." I'm not aware of any dependencies among the packages in that list, so I don't think its placement matters, and being on bottom is fine. Is that what you were thinking, too?

That's relevant when dependent projects get generated by the tool. So if STJ had dependencies which weren't already included in SBRP, the tool would generate projects for those. The whole set of those package names need to be listed in Build.props in dependency order.

PR #768, opened 2023-08-18
Add Newtonsoft.Json.13.0.1

mthalman:
After the [upgrade of Newtonsoft.Json to 13.0.3 in source-build-externals](https://github.com/dotnet/source-build-externals/pull/189), we need the previous version of 13.0.1 to be available in SBRP so that repos can still reference 13.0.1.


mthalman:
The build is failing because the project is not configured with the signing key:

```
artifacts/source-build/self/package-cache/microsoft.dotnet.arcade.sdk/8.0.0-beta.23411.1/tools/StrongName.targets(89,5): error : (NETCORE_ENGINEERING_TELEMETRY=Build) PublicKey, PublicKeyToken and AssemblyOriginatorKeyFile must be specified
```

But there is no .snk file published for [Newtonsoft.Json](https://github.com/JamesNK/Newtonsoft.Json) that we can include in this repo.

So this is blocked at this point. I'm going to revert https://github.com/dotnet/source-build-externals/pull/189 until this can be resolved.

MichaelSimons:
I think we want to get repo owners to upgrade so that everyone is on the same version - 13.0.3.  We should avoid adding non-msft reference packages.

mthalman:
Closing based on https://github.com/dotnet/source-build-reference-packages/pull/768#issuecomment-1686638764

PR #773, opened 2023-08-30
Add System.ComponentModel.Composition.4.5.0

mthalman:
Adds a ref package for System.ComponentModel.Composition.4.5.0. Necessary for https://github.com/microsoft/vstest/pull/4675.


MichaelSimons:
I don't think this should be added to SBRP because if I understand correctly this is a repo only requirement which would not be utilized for the product source-build.  This should be added as an allowed prebuilt.

PR #774, opened 2023-09-01
Manually update Microsoft.Private.Intellisense to 8.0.0-preview-20230828.1

carlossanlop:
As suggested in this comment: https://github.com/dotnet/runtime/pull/91285#issuecomment-1698593761 I ran:
```
generate.cmd/sh -p Microsoft.Private.Intellisense,8.0.0-preview-20230828.1 -t text
```

This is the workaround for https://github.com/dotnet/runtime/issues/77004

@ViktorHofer @mthalman @MichaelSimons please review to unblock https://github.com/dotnet/runtime/pull/91285


mthalman (commented):
Since runtime is the only repo referencing this package that participates in source-build, the old version can be deleted.

Also, as stated in the [workflow](https://github.com/dotnet/source-build-reference-packages#workflow), the new project should be added to Build.props.

ViktorHofer:
It's not just runtime. At least winforms uses that package as well. But they should also upgrade to the new version as part of this. Therefore deleting the old version makes sense.

ViktorHofer (approved):
Consider deleting the old version

carlossanlop:
@mthalman @ViktorHofer can I delete the old version in a separate PR? I first need to ensure that the other repos that consume the package have already upgraded to 20230828.1, otherwise we would end up with the same problem we're trying to fix, I am assuming.

carlossanlop:
I created https://github.com/dotnet/runtime/issues/91475 to track the deletion request and assigned it to myself, so that we can unblock this PR. @mthalman would you mind merging this?

mthalman:
> I created [dotnet/source-build#3611](https://github.com/dotnet/source-build/issues/3611) to track the deletion request and assigned it to myself, so that we can unblock this PR. @mthalman would you mind merging this?

Even without deleting the old version, this can't be merged until Build.props file is updated.

carlossanlop:
Package 7.0.0-preview-20221010.1 is consumed by .NET 7.0. Would it cause source-build problems in that .NET version if I delete that folder?

mthalman:
> Package 7.0.0-preview-20221010.1 is consumed by .NET 7.0. Would it cause source-build problems in that .NET version if I delete that folder?

No. SBRP is only used for .NET 8+.

carlossanlop:
Ok thanks for confirming. I deleted the old code and also added the missing item in Build.props.

mthalman (approved):

PR #778, opened 2023-09-07
Add targeting packs for 6.0.2

mthalman:
Necessary to support BrowserRefresh: https://github.com/dotnet/sdk/pull/35190


MichaelSimons (commented):
All xml doc files should be excluded from these targeting packs.  They add significant bloat to the repo/resulting packages.  These targeting packs are not intended for developer use and have been removed from existing packs.

MichaelSimons (approved):

PR #780, opened 2023-09-12
Update self-reference for RC1 release

MichaelSimons:

PR #782, opened 2023-09-13
[main] Update self-reference for RC1 release

MichaelSimons:
Port https://github.com/dotnet/source-build-reference-packages/pull/780 to main.

### DIFF
--- a/eng/Version.Details.xml
+++ b/eng/Version.Details.xml
@@ -6,9 +6,9 @@
       <Sha>33edde07d61cf7606d76ada765335fb81f1cbb71</Sha>
       <SourceBuild RepoName="source-build-externals" ManagedOnly="true" />
     </Dependency>
-    <Dependency Name="Microsoft.SourceBuild.Intermediate.source-build-reference-packages" Version="8.0.0-alpha.1.23368.2">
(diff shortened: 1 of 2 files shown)


eng/Version.Details.xml
```diff
@@ -6,9 +6,9 @@
       <Sha>33edde07d61cf7606d76ada765335fb81f1cbb71</Sha>
       <SourceBuild RepoName="source-build-externals" ManagedOnly="true" />
     </Dependency>
-    <Dependency Name="Microsoft.SourceBuild.Intermediate.source-build-reference-packages" Version="8.0.0-alpha.1.23368.2">
+    <Dependency Name="Microsoft.SourceBuild.Intermediate.source-build-reference-packages" Version="8.0.0-alpha.1.23414.1">
       <Uri>https://github.com/dotnet/source-build-reference-packages</Uri>
-      <Sha>529fbc2aad419d0c1551d6685cc68be33e62a996</Sha>
+      <Sha>45f5554ac6813e1e155c47fd80ec146c684f65e8</Sha>
```
mthalman, line 11: This should probably be updated to reflect the package from the new 9.0 channel.

MichaelSimons: No, it shouldn't until there is a release from main.  Main is or will be building soon with the rc1 SDK and this is the version it shipped with.  This would cause a type of incoherency.

mthalman (approved):

PR #794, opened 2023-09-29
Add SBRP attribute

ellahathaway:
Fixes https://github.com/dotnet/source-build/issues/3639

This PR adds the following attribute to assemblies originating from SBRP: `[assembly: AssemblyInformationalVersion("<version> originated from source-build-reference-packages")]`. This PR also includes logic to add this attribute as part of packageSourceGenerator.


src/referencePackages/src/humanizer.core/2.2.0/lib/netstandard1.0/Humanizer.cs
```diff
@@ -18,6 +18,7 @@
 [assembly: System.Reflection.AssemblyProduct("Humanizer (netstandard1.0)")]
 [assembly: System.Reflection.AssemblyTitle("Humanizer")]
 [assembly: System.Reflection.AssemblyVersionAttribute("2.2.0.0")]
+[assembly: System.Reflection.AssemblyInformationalVersion("2.2.0.0 originated from source-build-reference-packages")]
```
mthalman, line 21: This conflicts with the `AssemblyInformationalVersion` attribute on line 17. This is seemingly the case for all packages. Rather than adding a new attribute, I think the existing attribute will need to be modified.

src/packageSourceGenerator/PackageSourceGeneratorTask/AddSbrpAttribute.cs
```diff
@@ -0,0 +1,61 @@
...
+using System;
+using System.IO;
+using System.Linq;
+using System.Text.RegularExpressions;
+using Microsoft.Build.Framework;
+using Microsoft.Build.Utilities;
+
+namespace Microsoft.DotNet.SourceBuild.Tasks
+{
+    public partial class AddSbrpAttribute : Task
```
MichaelSimons, line 13: Naming question.  What do you envision as the scope of this task?  If we needed to add additional tweaks to the source, would this task be expanded or would these be another task added?  From a performance aspect, the file IO would be expensive to utilize multiple tasks.  With this in mind, what do you think about the name of this task?

src/referencePackages/src/humanizer.core/2.2.0/lib/netstandard1.0/Humanizer.cs
ellahathaway: Should the text "2.2.0+Branch.master.Sha.b30550eed103a6970d8465fe7c5c16300b70be81" be kept with "originated from source-build-reference-packages" appended to the end of the string, or should I replace the first attribute on line 17 with the new SBRP attribute?

src/packageSourceGenerator/PackageSourceGeneratorTask/AddSbrpAttribute.cs
```diff
@@ -0,0 +1,61 @@
...
+
+            Match versionMatch = GetVersionRegex().Match(specContent);
+
+            if (versionMatch.Success)
+            {
+                string version = versionMatch.Groups["version"].Value;
+
+                if (!GetSbrpAttributeRegex().IsMatch(specContent))
+                {
+                    string sbrpAttribute = $"[assembly: System.Reflection.AssemblyInformationalVersion(\"{version} originated from source-build-reference-packages\")]";
```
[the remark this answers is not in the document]
ellahathaway: Using AssemblyMetadataAttribute could work. It's not as descriptive as saying "originated from SBRP", but it still provides the necessary information.

src/packageSourceGenerator/PackageSourceGeneratorTask/AddSbrpAttribute.cs
ellahathaway: If we need to add additional attributes, I can see how running this task multiple times would be expensive. I'll change the name and adjust the task to have a parameter that is a list of attributes to add. This should eliminate needing to read and write the file multiple times.

src/packageSourceGenerator/PackageSourceGeneratorTask/AddSbrpAttribute.cs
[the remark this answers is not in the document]
mthalman: The parameter values are arbitrary and can be anything we want. So we can be as descriptive as necessary.

src/referencePackages/src/humanizer.core/2.2.0/lib/netstandard1.0/Humanizer.cs
ellahathaway: Addressed in this commit: https://github.com/dotnet/source-build-reference-packages/pull/794/commits/95dae831be4d21f4443766df377f7ee554ff01b8

src/packageSourceGenerator/PackageSourceGeneratorTask/AddSbrpAttribute.cs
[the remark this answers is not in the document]
ellahathaway: Addressed in this commit: https://github.com/dotnet/source-build-reference-packages/pull/794/commits/95dae831be4d21f4443766df377f7ee554ff01b8

src/packageSourceGenerator/PackageSourceGeneratorTask/AddSbrpAttribute.cs
```diff
@@ -0,0 +1,61 @@
...
+                }
+            }
+            return true;
+        }
+
+        private bool AddSbrpAttributeToSpec(string CSharpPath) {
+
+            string specContent = File.ReadAllText(CSharpPath!);
+
+            Match versionMatch = GetVersionRegex().Match(specContent);
```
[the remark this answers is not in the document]
ellahathaway: Refactored in this commit, which removed checking of the version attribute: https://github.com/dotnet/source-build-reference-packages/pull/794/commits/b6e2ed619dbe94e19bd75d42a170a22809445027

src/packageSourceGenerator/PackageSourceGeneratorTask/AddSbrpAttribute.cs
ellahathaway: Refactored in this commit to add arbitrary attributes: https://github.com/dotnet/source-build-reference-packages/pull/794/commits/b6e2ed619dbe94e19bd75d42a170a22809445027. This allows the scope of the task to be wider and eliminates writing to the file more than once.

ellahathaway:
Not sure why the Windows tests are failing: https://dev.azure.com/dnceng-public/public/_build/results?buildId=432336&view=logs&j=0ee6d9ba-6a42-510f-1190-caf8c1ddefec&t=03a2f049-f4d3-5343-2a38-569f0ad86e60&l=41

ViktorHofer (commented):
I'm not sure why we need to add this item to the generated compile source. Why can't we just define that attribute in a separate C# file and pass it to the compiler? Doing that will avoid these 600+ file changes and keep the logic contained in a single place.

EDIT: Even better, just add this this attribute via an msbuild item here: https://github.com/dotnet/source-build-reference-packages/blob/main/src/referencePackages/Directory.Build.props

```
    <AssemblyAttribute Include="System.Reflection.AssemblyMetadata">
      <_Parameter1>source</_Parameter1>
      <_Parameter2>source-build-reference-packages</_Parameter2>
    </AssemblyAttribute>
```

src/packageSourceGenerator/PackageSourceGeneratorTask/AddAttributes.cs
```diff
@@ -0,0 +1,65 @@
+// Licensed to the .NET Foundation under one or more agreements.
```
ViktorHofer, line 1: Delete this file

src/referencePackages/Directory.Build.props
```diff
@@ -1,6 +1,11 @@
 <Project>
 
   <Import Project="..\..\Directory.Build.props" />
+  
+  <ItemGroup>
+    <!-- Attributes for all Reference Packages -->
+    <AssemblyMetadata Include="source" Value="source-build-reference-packages" />
```
ViktorHofer, line 7: I don't think that's the correct usage. Please double check that works (open a compiled assembly with ilspy / ildasm).

ViktorHofer: Here's an example: https://github.com/dotnet/runtime/blob/53830e2608467ed6c2be38a782cc07803d960f53/eng/versioning.targets#L12-L15

ellahathaway: https://learn.microsoft.com/en-us/visualstudio/msbuild/common-msbuild-project-items?view=vs-2022#assemblymetadata

https://github.com/dotnet/docs/blob/main/docs/core/project-sdk/msbuild-props.md#assemblymetadata

These documents suggest that this usage is correct.

ViktorHofer: Interesting, I didn't know that.

MichaelSimons, line 7: I think we should add this to the targeting packs as well.  We shouldn't be shipping any of them.  The text-only packages  should not get this (just as you are proposing)

ViktorHofer: Agreed, makes sense. Maybe it's fine to just duplicate this logic in both Directory.Build.props files. It won't really change anytime in the future (hopefully).

ellahathaway:
After testing [the proposed implementation change](https://github.com/dotnet/source-build-reference-packages/pull/794#pullrequestreview-1667194392), I found that the rebuilt assemblies did not include the attribute. To investigate this, I created a "dummy" dotnet app with the same code to inject the attribute and found that the attribute was added to the assemblies as intended, indicating that the proposed changes do inject the attribute as intended, but something in the build of SBRP was causing the attribute to not appear.

I discovered that this issue is the result of setting GenerateAssemblyInfo to false. As described in [this issue](https://github.com/dotnet/project-system/issues/3934), when GenerateAssemblyInfo is false, the project file and the project properties become out of sync. I tested this with my "dummy" dotnet app by setting "GenerateAssemblyInfo" to false and verified that the attribute does not appear in the built assembly.

However, setting GenerateAssemblyInfo to true in SBRP caused a myriad of issues, including complaints about duplicate attributes.

I have two proposed changes to address this issue:

1) Set GenerateAssemblyInfo to true but don't regenerate the duplicate assemblies. [John Koerner - Dealing with Duplicate Assembly Attributes in .Net Core](https://johnkoerner.com/csharp/dealing-with-duplicate-attribute-errors-in-net-core/). I don't feel that this is a good long-term solution. Doing this means that future developers would have to manually set duplicate attributes to false as they are added.

2) Manually inject the attribute as originally done.

MichaelSimons:
@ellahathaway - Nice discovery and analysis of what is happening.

1. Are you referring to assembly attributes that don't exist today but could be defined in the future?  Even if this happened, isn't this as simple as adding a single property in the root referencePackages Directory.Build.props?  Without understanding this drawback fully, this feels like the most compelling option to me.
2. Keep in mind that we do have a code generator so while it would touch a lot of files it would be done via automation.  That being said, the GenAPI logic is what generates these attributes and it doesn't feel like the best choice for us to have what ever logic is necessary to specify these via the project file.

ViktorHofer:
We don't want to generate an AssemblyInfo.cs file as GenAPI already emits all the assembly metadata into the compiled source. I think a better solution would be to just generate that C# code ourselves (basically what the `GenerateAssemblyInfo` target/task is doing) or even just hardcode it as a C# file and add it as a compile input.

MichaelSimons:
> We don't want to generate an AssemblyInfo.cs file as GenAPI already emits all the assembly metadata into the compiled source.

I didn't know if there would be enough knobs as to only generate the AssemblyMetadataAttribute.

ellahathaway:
> even just hardcode it as a C# file and add it as a compile input.

I think this is a good approach. I'll work on add this.

src/targetPacks/Directory.Build.props
```diff
@@ -1,6 +1,11 @@
 <Project>
   <Import Project="..\..\Directory.Build.props" />
 
+    <ItemGroup>
```
MichaelSimons, line 4: This item group is indented too far.  msbuild artifacts use an indent of 2 spaces instead of 4.

src/referencePackages/Directory.Build.targets
```diff
@@ -120,6 +120,11 @@
 
   <Target Name="CopyBuildOutputToTempOutput"
           AfterTargets="CopyFilesToOutputDirectory">
+
+    <ItemGroup>
```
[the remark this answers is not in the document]
ellahathaway: This removes the .cs file from the Compile sources. If it's included here, the output gets copied to the parent directory of the .cs file as part of this target.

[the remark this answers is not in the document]
ViktorHofer: Yes, that's the "binplacing" target. It currently depends on the a single compile item.

To make this more robust, it would be better to change all usages of the `Compile` item to another item that only contains the actual compile item.

So: L129 & 133 shouldn't use `@(Compile)` anymore but another item.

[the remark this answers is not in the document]
ViktorHofer: so something like this:

```
<ItemGroup>
  <CompileForRelativePath Include="@(Compile)" />
  <CompileForRelativePath Remove="$(RepoRoot)src\AssemblyInfo.cs" />
</ItemGroup>
```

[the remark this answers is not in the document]
ellahathaway: Thanks for the suggestion! Added here: https://github.com/dotnet/source-build-reference-packages/pull/794/commits/60266f871372130df00b5cf104002e5951aca8b1#diff-281c53fe4841f457012b2ddead53ad727b69e27c1f6d4911f2ada3042d035af0R125

src/targetPacks/Directory.Build.props
ellahathaway: Fixed here: https://github.com/dotnet/source-build-reference-packages/pull/794/commits/60266f871372130df00b5cf104002e5951aca8b1#diff-d8109394bde86e6ff38bbfae32c914420f607657372dd94f27448aa406a6e9d8R4

src/AssemblyInfo.cs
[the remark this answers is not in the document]
ellahathaway: Renamed in this commit: https://github.com/dotnet/source-build-reference-packages/pull/794/commits/60266f871372130df00b5cf104002e5951aca8b1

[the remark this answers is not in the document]
ellahathaway: Added in this commit: https://github.com/dotnet/source-build-reference-packages/pull/794/commits/60266f871372130df00b5cf104002e5951aca8b1

ViktorHofer (approved):

ViktorHofer:
I assume you verified that the assembly attribute is now present? Should we add a test for it (post-merge)?

MichaelSimons:
> I assume you verified that the assembly attribute is now present? Should we add a test for it (post-merge)?

Testing of this attribute was already tracked in https://github.com/dotnet/source-build/issues/2804 which will be one of the next items Ella will be working on.

MichaelSimons (approved):

src/packageSourceGenerator/PackageSourceGeneratorTask/AddAttributes.cs
```diff
@@ -0,0 +1,65 @@
...
+                }
+                else
+                {
+                    Log.LogMessage(MessageImportance.Low, $"Attributes already exist in {path}");
+                }
+            }
+            return true;
+        }
+
+        private bool AddAttributesToFile(string CSharpPath) {
```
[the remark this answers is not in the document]
mthalman: Disregard. From an older review.

src/packageSourceGenerator/PackageSourceGeneratorTask/AddAttributes.cs
```diff
@@ -0,0 +1,65 @@
...
+
+            string fileContent = File.ReadAllText(CSharpPath!);
+
+            string[] attributesNotInFile = Attributes
+                .Where(attribute => !fileContent.Contains(attribute))
+                .ToArray();
+
+            string attributesToAppend = string.Join(Environment.NewLine, attributesNotInFile);
+            string replacement = Environment.NewLine + attributesToAppend + "\n[assembly:";
```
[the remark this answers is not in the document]
mthalman: Disregard. From an older review.

ellahathaway:
> I assume you verified that the assembly attribute is now present? Should we add a test for it (post-merge)?

I did check to verify that the assembly attribute is now present!

PR #805, opened 2023-10-20
Update Newtonsoft 13.0.1 referenced to 13.0.3

MichaelSimons:
Update Newtonsoft 13.0.1 referenced to 13.0.3 to prevent online only prebuilts.  This is fallout from https://github.com/dotnet/source-build-externals/pull/195.

### DIFF
--- a/src/referencePackages/src/microsoft.extensions.dependencymodel/3.0.0/Microsoft.Extensions.DependencyModel.3.0.0.csproj
+++ b/src/referencePackages/src/microsoft.extensions.dependencymodel/3.0.0/Microsoft.Extensions.DependencyModel.3.0.0.csproj
@@ -9,7 +9,7 @@
 
   <ItemGroup Condition="'$(TargetFramework)' == 'netstandard1.3'">
     <!-- Manually updated version from 9.0.1 to address https://github.com/advisories/GHSA-5crp-9r3c-p9vr -->
-    <PackageReference Include="Newtonsoft.Json" Version="13.0.1" />
+    <PackageReference Include="Newtonsoft.Json" Version="13.0.3" />
     <PackageReference Include="System.AppContext" Version="4.1.0" />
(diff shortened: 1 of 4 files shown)


MichaelSimons:
FYI, in 9.0 we now have a check to fail online builds if they have prebuilts.  This would have caught this.

mmitche (approved):

mthalman (approved):

ViktorHofer:
The failure says there's still a Newtonsoft.Json/13.0.1 in the dependency graph. Do you need help with identifying the project that brings it in? The project.assets.json file should tell.

MichaelSimons:
> The failure says there's still a Newtonsoft.Json/13.0.1 in the dependency graph. Do you need help with identifying the project that brings it in? The project.assets.json file should tell.

This is a nuget restore source issue in online builds, not a problem with the actual SBRP packages.  This wouldn't be an issue if we have proper project dependencies and no n-1 SBRP packages around.

oleksandr-didyk (approved):

eng/SourceBuildPrebuiltBaseline.xml
```diff
@@ -7,6 +7,10 @@
     <UsagePattern IdentityGlob="runtime.linux-x64.Microsoft.NETCore.ILAsm/*" />
     <UsagePattern IdentityGlob="runtime.linux-x64.Microsoft.NETCore.ILDAsm/*" />
 
+    <!-- Necessary to work around NuGet non-determinism with online builds/feeds. Can be removed once 
+    dependencies on it are removed from the previous-source-built packages. -->
+    <UsagePattern IdentityGlob="newtonsoft.json/13.0.1" />
```
ViktorHofer, line 12:
```suggestion
    <UsagePattern IdentityGlob="Newtonsoft.Json/13.0.1" />
```

PR #812, opened 2023-10-26
Add more ref packages

dougbu:
- add Microsoft.Extensions.FileProviders.Abstractions.6.0.0
- add Microsoft.Extensions.FileSystemGlobbing.6.0.0

### DIFF
--- a/eng/Build.props
+++ b/eng/Build.props
@@ -31,6 +31,7 @@
     <DependencyPackageProjects Include="$(RepoRoot)src\referencePackages\src\**\NuGet.Packaging.6.7.0.csproj" />
     <DependencyPackageProjects Include="$(RepoRoot)src\referencePackages\src\**\NuGet.Protocol.6.7.0.csproj" />
     <DependencyPackageProjects Include="$(RepoRoot)src\referencePackages\src\**\NuGet.Credentials.6.7.0.csproj" />
+
     <DependencyPackageProjects Include="$(RepoRoot)src\referencePackages\src\**\Microsoft.Extensions.FileProviders.Abstractions.6.0.0.csproj" />
     <DependencyPackageProjects Include="$(RepoRoot)src\referencePackages\src\**\Microsoft.Extensions.FileSystemGlobbing.6.0.0.csproj" />
   </ItemGroup>
--- a/eng/Build.props
+++ b/eng/Build.props
@@ -31,6 +31,8 @@
     <DependencyPackageProjects Include="$(RepoRoot)src\referencePackages\src\**\NuGet.Packaging.6.7.0.csproj" />
     <DependencyPackageProjects Include="$(RepoRoot)src\referencePackages\src\**\NuGet.Protocol.6.7.0.csproj" />
     <DependencyPackageProjects Include="$(RepoRoot)src\referencePackages\src\**\NuGet.Credentials.6.7.0.csproj" />
+    <DependencyPackageProjects Include="$(RepoRoot)src\referencePackages\src\**\Microsoft.Extensions.FileProviders.Abstractions.6.0.0.csproj" />
+    <DependencyPackageProjects Include="$(RepoRoot)src\referencePackages\src\**\Microsoft.Extensions.FileSystemGlobbing.6.0.0.csproj" />
   </ItemGroup>
 
   <ItemGroup Condition="'$(BuildDependencyPackageProjects)' == 'true'">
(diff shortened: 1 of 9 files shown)


dougbu:
I have no idea whether this will build. Locally (in WSL) for me, `./build.sh` always causes `dotnet` to exit shortly after restoring packages w/ a `137` exit code. Is that a known issue w/ a known workaround❔

dougbu:
see https://github.com/dotnet/arcade/pull/14003/checks?check_run_id=18098774810 for work this enables

mthalman:
> I have no idea whether this will build. Locally (in WSL) for me, `./build.sh` always causes `dotnet` to exit shortly after restoring packages w/ a `137` exit code. Is that a known issue w/ a known workaround❔

You need to use the `-sb` switch as documented here: https://github.com/dotnet/source-build-reference-packages#building

MichaelSimons:
> Otherwise, these new projects don't end up getting built.

More precisely, they may not get built in the correct order therefore causing prebuilts.

dougbu:
> In the [workflow instructions](https://github.com/dotnet/source-build-reference-packages#workflow), it states that the eng/build.props file needs to be updated to reference the new packages. Otherwise, these new projects don't end up getting built.

Couldn't find anything worth adding to the file. I believe the workflow instructions were written before eng/Build.props was written to use `$(RepoRoot)src\referencePackages\src\**\*.csproj`.

MichaelSimons:
> > In the [workflow instructions](https://github.com/dotnet/source-build-reference-packages#workflow), it states that the eng/build.props file needs to be updated to reference the new packages. Otherwise, these new projects don't end up getting built.
> 
> Couldn't find anything worth adding to the file. I believe the workflow instructions were written before eng/Build.props was written to use `$(RepoRoot)src\referencePackages\src\**\*.csproj`.

They need to be added to this section - https://github.com/dotnet/source-build-reference-packages/blob/main/eng/Build.props#L13

dougbu:
> > I have no idea whether this will build. Locally (in WSL) for me, `./build.sh` always causes `dotnet` to exit shortly after restoring packages w/ a `137` exit code. Is that a known issue w/ a known workaround❔
> 
> You need to use the `-sb` switch as documented here: https://github.com/dotnet/source-build-reference-packages#building

I used that switch but it didn't avoid the `137` exit codes. Might be specific to my WSL environment.

MichaelSimons:
> > > I have no idea whether this will build. Locally (in WSL) for me, `./build.sh` always causes `dotnet` to exit shortly after restoring packages w/ a `137` exit code. Is that a known issue w/ a known workaround❔
> > 
> > 
> > You need to use the `-sb` switch as documented here: https://github.com/dotnet/source-build-reference-packages#building
> 
> I used that switch but it didn't avoid the `137` exit codes. Might be specific to my WSL environment.

Do you have the SDK's native dependencies installed?  Do you use the environment to build other repos in?

dougbu:
> They need to be added to this section - https://github.com/dotnet/source-build-reference-packages/blob/main/eng/Build.props#L13

I'm confused. I didn't add anything that another project in the repo depends on.

dougbu:
> Do you have the SDK's native dependencies installed? Do you use the environment to build other repos in?

Yes and yes but it's probably not too important to worry about my specific build issues.

MichaelSimons:
> I'm confused. I didn't add anything that another project in the repo depends on.

I apologize for the wording.  All new projects need to be added here regardless if they are referenced or not.  I will update the wording in a PR I am working on to update the readme per some suggestions you sent me offline.

dougbu:
> > I'm confused. I didn't add anything that another project in the repo depends on.
> 
> &hellip; All new projects need to be added here regardless if they are referenced or not. &hellip;

:ok: is this true for both projects I'm adding or only the brand-new microsoft.extensions.fileproviders.abstractions folder❔

dougbu:
> > … All new projects need to be added here regardless if they are referenced or not. …
> 
> 🆗 is this true for both projects I'm adding or only the brand-new microsoft.extensions.fileproviders.abstractions folder❔

/fyi I checked binary logs from the validation build. The new packages are generated in Build.binlog and sourcebuild.binlog but (as expected) not mentioned in sourcebuild-dependency-projects.binlog.

I'll add both projects to the referenced Build.props section. Feel free to back an unnecessary line out.

MichaelSimons:
> /fyi I checked binary logs from the validation build. The new packages are generated in Build.binlog and sourcebuild.binlog but (as expected) not mentioned in sourcebuild-dependency-projects.binlog.

image

MichaelSimons (approved):

MichaelSimons:
Thanks for adding these Doug and providing feedback to improve the UX!

PR #816, opened 2023-11-01
Misc improvements to workflow documentation

MichaelSimons:
I broke these changes into two commits.  The first one contains the content modifications.  The second one is just shortening some long lines.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,8 +54,14 @@ regenerate the older version to see what customizations to the generated code we
 #### Workflow
 
 * Generate reference package and its depencencies running the `./generate.sh --package <package>,<version>` script.
-* Revert any changes to packages that already existed in the repository.
-This is likely reverting a necessary source modification to the generated code.
+* Inspect any changes to packages that already existed in the repository. There are two reasons why previously
+generated packages show changes when being regenerated.
+    1. The package contains intentional code modifications on top of the generated code. This may be upgrading a
+    project reference to address a CVE or code fixups because the generate tooling does not support a scenario.
+    When this occurs, there should be code comments explaining why the code modification was made. If this is
+    the case, the changes to the existing package should be reverted.
+    2. The generate tooling has changed since the last time this package was generated. The new changes should
+    be considered better/correct and should be committed.
 * Add `DependencyPackageProjects` for all new projects in the
 [eng/Build.props](https://github.com/dotnet/source-build-reference-packages/blob/main/eng/Build.props#L9)
 in the correct dependency order.
--- a/README.md
+++ b/README.md
@@ -48,22 +48,21 @@ to address this manual step.
 
 The tooling does not handle all situations and sometimes the generated code will need manual tweeks to get it to compile.
 If this occurs when generating a newer version of an existing package, it can be helpful to regenerate the older version
-to see what changes were made.
-
-The tooling has evolved over time and therefore when generating new packages, you may see edits made to existing packages.
-This is because the new package has a dependency on an existing package, it was regenerated and changes were detected from
-when it was originally generated.
+to see what customizations to the generated code were made.
 
 #### Workflow
 
 * Generate reference package and its depencencies running the `./generate.sh --package <package>,<version>` script.
-* Revert changes for packages that were already existed in the repository.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -48,22 +48,21 @@ to address this manual step.
...
-This is because the new package has a dependency on an existing package, it was regenerated and changes were detected from
-when it was originally generated.
+to see what customizations to the generated code were made.
 
 #### Workflow
 
 * Generate reference package and its depencencies running the `./generate.sh --package <package>,<version>` script.
-* Revert changes for packages that were already existed in the repository.
-* Add `DependencyPackageProjects` for all new projects in the [eng/Build.props](https://github.com/dotnet/source-build-reference-packages/blob/main/eng/Build.props#L9)
+* Revert any changes to packages that already existed in the repository.
```
mthalman, line 57: I feel like we should have a test which validates the packages are up-to-date wrt the generator, specifically for GenAPI dependency flow PRs. It's really annoying to have to revert these changes.

README.md
```diff
@@ -48,22 +48,21 @@ to address this manual step.
...
+This is likely reverting a necessary source modification to the generated code.
```
mthalman, line 58: This sentence seems at odds with the previous one. I think it will cause confusion. This makes it sound like it's a source modification that's needed but they should revert it -- which is confusing.

README.md
MichaelSimons: How often are you seeing changes that aren't intentional customizations now that everything has been regenerated with the genapi based tooling?  Regenerating the code is not feasible at this point for two reasons.

1. The tooling does not support re-generation scenarios today.  There are some tracking issues that prevent this today.
2. There are a number of intentional changes - e.g. explicit package upgrades to address CVEs as well as code modifications that genAPI does not support.

mthalman: I had to revert modifications to at least 2 packages I can remember for my changes in https://github.com/dotnet/source-build-reference-packages/pull/815.

MichaelSimons: Were they necessary changes that were manually applied to the packages to address a CVE or to get the code to compile or were they changes because genAPI was "improved"?

mthalman: I think it was just "improved".

README.md
MichaelSimons: Reworded to:

```
Inspect any changes to packages that already existed in the repository. There are two reasons why previously
generated packages show changes when being regenerated.
    1. The package contains intentional code modifications on top of the generated code. This may be upgrading a
    project reference to address a CVE or code fixups because the generate tooling does not support a scenario.
    When this occurs, there should be code comments explaining why the code modification was made. If this is
    the case, the changes to the existing package should be reverted.
    2. The generate tooling has changed since the last time this package was generated. The new changes should
    be considered better/correct and should be committed.
```

README.md
MichaelSimons: If you feel like regenerating the packages on dependency flow would improve the workflow, please log an issue for this.

README.md
```diff
@@ -43,34 +44,40 @@ When generating a package(s), the tooling will detect and generate all dependent
...
-* Revert changes for packages that were already existed in the repository.
-* Add `DependencyPackageProjects` for all new projects in the [eng/Build.props](https://github.com/dotnet/source-build-reference-packages/blob/main/eng/Build.props#L9)
+* Inspect any changes to packages that already existed in the repository. There are two reasons why previously
+generated packages show changes when being regenerated.
+    1. The package contains intentional code modifications on top of the generated code. This may be upgrading a
+    project reference to address a CVE or code fixups because the generate tooling does not support a scenario.
+    When this occurs, there should be code comments explaining why the code modification was made. If this is
+    the case, the changes to the existing package should be reverted.
+    2. The generate tooling has changed since the last time this package was generated. The new changes should
+    be considered better/correct and should be committed.
```
mthalman, line 64: This may end up being difficult for newbies to differentiate. We should strive to add comments to the source whenever making manual modifications so they are easy to spot.

mthalman (approved):

README.md
MichaelSimons: Agreed.  Hopefully the last step I added will help ensure comments are added.

oleksandr-didyk (approved):

PR #817, opened 2023-11-02
SBRP validation tests

ellahathaway:
Closes https://github.com/dotnet/source-build/issues/2804

Wrote tests to
1) Check for SBRP attribute in reference packages
2) Check all packages for signature

The tests fail if a reference package is missing the SBRP attribute or if any package is signed


eng/Build.props
```diff
@@ -6,6 +6,7 @@
 
   <ItemGroup Condition="'$(Test)' == 'true'">
     <ProjectToBuild Include="$(RepoRoot)tests\GenerateScriptTests\GenerateScript.Tests.csproj" Test="$(Test)" />
+    <ProjectToBuild Include="$(RepoRoot)tests\ValidateSbrpTests\ValidateSbrp.Tests.csproj" Test="$(Test)" />
```
mthalman, line 9: We can simplify things by having just one test project (should probably be renamed to something generic like `SbrpTests`).

tests/ValidateSbrpTests/ValidateSbrpTests.cs
```diff
@@ -0,0 +1,170 @@
...
+
+namespace ValidateSbrpTests;
+
+public class ValidateSbrpTests
+{
+    private string[] Packages { get; set; }
+
+    private const string SbrpAttributeType = "System.Reflection.AssemblyMetadataAttribute";
+
+    private const string VerionPattern = @"(\.\d)+([\-\w])*";
```
mthalman, line 29: nit: spelling

```suggestion
    private const string VersionPattern = @"(\.\d)+([\-\w])*";
```

tests/ValidateSbrpTests/ValidateSbrpTests.cs
```diff
@@ -0,0 +1,170 @@
...
+
+    private string RepoRoot { get; set; }
+
+    public ITestOutputHelper output { get; set; }
```
mthalman, line 33: The properties can just be defined as `static readonly` fields. The members should be grouped together of like kinds. So constants should be group together, and fields together, etc.

tests/ValidateSbrpTests/ValidateSbrpTests.cs
```diff
@@ -0,0 +1,170 @@
...
+        }
+
+        Packages = Directory.GetFiles(buildPackagesDirectory, "*.nupkg", SearchOption.AllDirectories);
+    }
+
+    [Fact]
+    public void CheckForSbrpAttribute()
+    {
+        string[] targetPacks = Directory.GetDirectories(Path.Combine(RepoRoot, "src/targetPacks/ILsrc")).Select(x => Path.GetFileName(x).ToLower()).ToArray();
+        string[] textOnlyPackages = Directory.GetDirectories(Path.Combine(RepoRoot, "src/textOnlyPackages/src")).Select(x => Path.GetFileName(x).ToLower()).ToArray();
```
mthalman, line 55: Consider breaking these apart into separate lines for readability.

```suggestion
        string[] targetPacks = Directory.GetDirectories(Path.Combine(RepoRoot, "src/targetPacks/ILsrc"))
            .Select(x => Path.GetFileName(x).ToLower())
            .ToArray();
        string[] textOnlyPackages = Directory.GetDirectories(Path.Combine(RepoRoot, "src/textOnlyPackages/src"))
            .Select(x => Path.GetFileName(x).ToLower())
            .ToArray();
```

tests/ValidateSbrpTests/ValidateSbrpTests.cs
```diff
@@ -0,0 +1,170 @@
...
+        
+        foreach (var package in Packages)
+        {
+            string packageName = Path.GetFileNameWithoutExtension(package).ToLower();
+            packageName = Regex.Replace(packageName, VerionPattern, string.Empty);
+            if (!targetPacks.Contains(packageName) && !textOnlyPackages.Contains(packageName))
+            {
+                filteredPackages.Add(package);
+            }
+        }
```
mthalman, line 67: This can be simplified with LINQ methods. Not only is it less code but it's more semantically correct, IMO. That's because it makes clear that the intent is to filter `Packages` through the use of the `Where` method. With the previous code, it requires more careful inspection that filtering is all it's doing.

```suggestion
        var filteredPackages = Packages
            .Where(package =>
            {
                string packageName = Path.GetFileNameWithoutExtension(package).ToLower();
                packageName = Regex.Replace(packageName, VerionPattern, string.Empty);
                return !targetPacks.Contains(packageName) && !textOnlyPackages.Contains(packageName);
            });
```

tests/ValidateSbrpTests/ValidateSbrpTests.cs
```diff
@@ -0,0 +1,170 @@
...
+            catch
+            {
+                throw new InvalidOperationException($"{SbrpAttributeType} is not formatted properly with a key, value pair.");
+            }
+        }
+
+        return false;
+    }
+
+    public async Task<bool> IsPackageSigned(string packagePath, PackageSignatureVerifier verifier, SignedPackageVerifierSettings settings)
```
mthalman, line 162: Convention is for async methods to have an `Async` suffix:

```suggestion
    public async Task<bool> IsPackageSignedAsync(string packagePath, PackageSignatureVerifier verifier, SignedPackageVerifierSettings settings)
```

tests/ValidateSbrpTests/ValidateSbrpTests.cs
```diff
@@ -0,0 +1,170 @@
...
+        {
+            bool isSigned = await IsPackageSigned(package, verifier, settings);
+            Assert.False(isSigned, $"{package} is signed. Signed packages are not allowed in source-build-reference-packages.");
+        }
+    }
+
+    private bool HasSbrpAttribute(MetadataReader reader)
+    {
+        return reader.CustomAttributes.Select(attrHandle => reader.GetCustomAttribute(attrHandle)).Any(attr => IsAttributeSbrp(reader, attr));
+    }
```
mthalman, line 119: Simplify with expression-bodied method and break apart into separate lines for readability:

```suggestion
    private bool HasSbrpAttribute(MetadataReader reader) =>
        reader.CustomAttributes
            .Select(attrHandle => reader.GetCustomAttribute(attrHandle))
            .Any(attr => IsAttributeSbrp(reader, attr));
```

tests/ValidateSbrpTests/ValidateSbrp.Tests.csproj
```diff
@@ -0,0 +1,16 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>$(NetCurrent)</TargetFramework>
+    <Nullable>enable</Nullable>
+    <NoWarn>$(NoWarn);CS8602</NoWarn>
```
MichaelSimons, line 6: I don't think these nullable related warnings should be suppressed rather they should be addressed in code.

FYI, you could define these as one property - <NoWarn>$(NoWarn);CS8602;CS8603</NoWarn>

tests/ValidateSbrpTests/ValidateSbrpTests.cs
```diff
@@ -0,0 +1,170 @@
...
+            {
+                Directory.Delete(tempDirectory, true);
+            }
+        }
+    }
+
+    [Fact]
+    public async Task CheckForSignature()
+    {
+        var trustProviders = new List<ISignatureVerificationProvider>()
```
MichaelSimons, line 101: Consider utilizing an array here versus the overhead of instantiating a list.  It also helps simplify the code a little.

```
ISignatureVerificationProvider[] trustProviders = [new SignatureTrustAndValidityVerificationProvider()];
var verifier = new PackageSignatureVerifier(trustProviders);
var settings = SignedPackageVerifierSettings.GetDefault();
```

tests/ValidateSbrpTests/ValidateSbrpTests.cs
```diff
@@ -0,0 +1,170 @@
...
+                throw new InvalidOperationException($"{SbrpAttributeType} is not formatted properly with a key, value pair.");
+            }
+        }
+
+        return false;
+    }
+
+    public async Task<bool> IsPackageSigned(string packagePath, PackageSignatureVerifier verifier, SignedPackageVerifierSettings settings)
+    {
+        using (var packageReader = new PackageArchiveReader(packagePath))
```
MichaelSimons, line 164: FYI You can define a using declaration without braces.  This is a usage scenario where that syntax is nice IMO.

```
using var packageReader = new PackageArchiveReader(packagePath);
var result = await verifier.VerifySignaturesAsync(packageReader, settings, CancellationToken.None);
return result.IsSigned;
```

tests/ValidateSbrpTests/ValidateSbrpTests.cs
```diff
@@ -0,0 +1,170 @@
...
+
+    public ITestOutputHelper output { get; set; }
+
+    public ValidateSbrpTests(ITestOutputHelper _output)
+    {
+        output = _output;
+
+        RepoRoot = Environment.CurrentDirectory.Substring(0, Environment.CurrentDirectory.IndexOf("artifacts"));
+
+        string buildPackagesDirectory = Path.Combine(RepoRoot, "artifacts/source-build/self/src/artifacts/packages/Release/Shipping");
```
MichaelSimons, line 41: This code makes an assumption that all builds will be release builds.  There can also be debug builds.

tests/ValidateSbrpTests/ValidateSbrpTests.cs
```diff
@@ -0,0 +1,170 @@
...
+
+        if (!Directory.Exists(buildPackagesDirectory))
+        {
+            throw new DirectoryNotFoundException($"Directory {buildPackagesDirectory} does not exist. Try building with './build.sh -sb' first.");
+        }
+
+        Packages = Directory.GetFiles(buildPackagesDirectory, "*.nupkg", SearchOption.AllDirectories);
```
MichaelSimons, line 48: Consider adding a check to ensure there are packages.  e.g. nonzero or maybe even more than 200 or whatever is close to the actual count.  If no packages are found, it looks like the tests would still pass.  Perhaps emitting how many packages are being tested would be beneficial.

tests/ValidateSbrpTests/ValidateSbrpTests.cs
```diff
@@ -0,0 +1,170 @@
...
+
+    private const string SbrpAttributeType = "System.Reflection.AssemblyMetadataAttribute";
+
+    private const string VerionPattern = @"(\.\d)+([\-\w])*";
+
+    private string RepoRoot { get; set; }
+
+    public ITestOutputHelper output { get; set; }
+
+    public ValidateSbrpTests(ITestOutputHelper _output)
```
MichaelSimons, line 35: The `_` prefix for parameter names is opposite of one of the standard style conventions used in .NET.  internal and private fields typically use the prefix.  This is a potential source of confusion.  https://github.com/dotnet/runtime/blob/main/docs/coding-guidelines/coding-style.md

tests/ValidateSbrpTests/ValidateSbrpTests.cs
ellahathaway: Out of curiosity, why is `private static readonly` preferred to `private const`?

My understanding is that const variables are implicitly static and defined at compile time, and static readonly variables are static and are defined at runtime. If `SbrpAttributeType` and `VersionPattern` are already initialized, why is static readonly preferred to const?

mthalman: I'm referring to the properties, not the existing `const` members. The comment range ended up including the `const` members because the properties surrounded them. Yes, where possible use `const` if its value is not determined at runtime.

tests/ValidateSbrpTests/ValidateSbrp.Tests.csproj
```diff
@@ -0,0 +1,16 @@
...
+    <NoWarn>$(NoWarn);CS8603</NoWarn>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="NuGet.Packaging" Version="$(NuGetPackagingVersion)" />
+    <PackageReference Include="System.IO.Compression" Version="4.3.0" />
+    <PackageReference Include="System.IO.Compression.ZipFile" Version="4.3.0" />
```
ViktorHofer, line 13: These old 4.3.0 packages shouldn't be referenced anymore these days. They will eventually be marked as deprecated. Curious, why did you bring them in? They shouldn't be needed anymore.

ellahathaway: I was experiencing compile errors for `ZipFile` (not being recognizined, needing to add the namespace, adding the namespace and it still giving the warnings -> added the package references).. adding the references fixed the issues. In hindsight, I think it's because I was using an old version of the sdk. Updating which sdk I was using fixed the issues I was having. Still a bit unsure of what fixed the issue I was having, but after creating a new container and reinstalling everything, I removed the package references and everything compiles as desired.

eng/Build.props
ellahathaway: Addressed in https://github.com/dotnet/source-build-reference-packages/pull/817/commits/ce1a70dc071c6a5adce2b2d0ca6285031ee12258

tests/ValidateSbrpTests/ValidateSbrpTests.cs
ellahathaway: Fixed in https://github.com/dotnet/source-build-reference-packages/pull/817/commits/ce1a70dc071c6a5adce2b2d0ca6285031ee12258

tests/ValidateSbrpTests/ValidateSbrpTests.cs
ellahathaway: Changed in https://github.com/dotnet/source-build-reference-packages/commit/ce1a70dc071c6a5adce2b2d0ca6285031ee12258

tests/ValidateSbrpTests/ValidateSbrpTests.cs
ellahathaway: Changed in https://github.com/dotnet/source-build-reference-packages/commit/ce1a70dc071c6a5adce2b2d0ca6285031ee12258

tests/ValidateSbrpTests/ValidateSbrpTests.cs
ellahathaway: Changed in https://github.com/dotnet/source-build-reference-packages/commit/ce1a70dc071c6a5adce2b2d0ca6285031ee12258

tests/ValidateSbrpTests/ValidateSbrpTests.cs
ellahathaway: Changed in https://github.com/dotnet/source-build-reference-packages/commit/ce1a70dc071c6a5adce2b2d0ca6285031ee12258

tests/ValidateSbrpTests/ValidateSbrp.Tests.csproj
ellahathaway: Removed in https://github.com/dotnet/source-build-reference-packages/commit/ce1a70dc071c6a5adce2b2d0ca6285031ee12258 and added null-forgiving operator to relevant lines in `DummyAttributeTypeProvider.cs`

tests/ValidateSbrpTests/ValidateSbrpTests.cs
ellahathaway: Fixed in https://github.com/dotnet/source-build-reference-packages/commit/ce1a70dc071c6a5adce2b2d0ca6285031ee12258

tests/ValidateSbrpTests/ValidateSbrpTests.cs
ellahathaway: Changed in https://github.com/dotnet/source-build-reference-packages/commit/ce1a70dc071c6a5adce2b2d0ca6285031ee12258

tests/ValidateSbrpTests/ValidateSbrpTests.cs
ellahathaway: Added checks in https://github.com/dotnet/source-build-reference-packages/commit/ce1a70dc071c6a5adce2b2d0ca6285031ee12258

tests/ValidateSbrpTests/ValidateSbrpTests.cs
ellahathaway: Changed in https://github.com/dotnet/source-build-reference-packages/commit/ce1a70dc071c6a5adce2b2d0ca6285031ee12258. Made into `Output` and applied similar changes to GenerateScriptTests.cs for consistency.

ellahathaway:
@mthalman @MichaelSimons SBRP Validation tests require investigating the built packages. The reason the build is failing in this PR is that the job for running the tests does not execute a build (this was not an issue for GenerateScriptTests.cs as it was independent of the build job).

Due to this, some refactoring must be done. Given that the tests are run on Windows and Linux, I'm wondering if we would consider running 3 jobs rather than 4. This would mean changing the pipeline.

The current pipeline is:
Job 1: Run SB managed build and publish artifacts
Job 2: Run tests for Windows
Job 3: Run tests for Linux
Job 4: Run SB complete

My proposed pipeline is:
Job 1: Build, run tests, and publish artifacts Windows
Job 2: Build, run tests, and publish artifacts for Linux
Job 3: Run SB complete

tests/SbrpTests/DummyAttributeTypeProvider.cs
```diff
@@ -6,25 +6,25 @@
...
 
-namespace ValidateSbrpTests;
+namespace SbrpTests;
 
 internal class DummyAttributeTypeProvider : ICustomAttributeTypeProvider<Type>
 {
     public static readonly DummyAttributeTypeProvider Instance = new();
 
-    public Type GetPrimitiveType(PrimitiveTypeCode typeCode) => default(Type);
+    public Type GetPrimitiveType(PrimitiveTypeCode typeCode) => default(Type)!;
```
mthalman, line 15: The `!` portion of this doesn't seem correct. `Type` is a `class`, so `default(Type)` will be `null`. The return type of the method is `Type`, which is non-nullable. So yes, adding `!` takes care of the compiler error but it doesn't make it correct since it will be null at runtime and break the nullability contract. The correct thing to do would be to change it from `ICustomAttributeTypeProvider<Type>` to `ICustomAttributeTypeProvider<Type?>`, and all the return types to be `Type?`. That will allow `null` or `default(Type)` to be returned.

tests/SbrpTests/Utilities.cs
```diff
@@ -0,0 +1,39 @@
...
+                    {
+                        throw new InvalidOperationException($"Environment variable '{Config.BuildTypeEnv}' must be set to either 'Debug' or 'Release'.");
+                    }
+                    break;
+                default:
+                    throw new InvalidOperationException($"Unknown environment variable '{parameter}'.");
+            }
+        }
+    }
+}
```
mthalman, line 39: nit: missing newline

tests/SbrpTests/ValidationTests.cs
```diff
@@ -18,71 +18,69 @@
...
-        string buildPackagesDirectory = Path.Combine(RepoRoot, "artifacts/source-build/self/src/artifacts/packages/Release/Shipping");
+        Utilities.ValidateConfigParameters(new string[] { Config.RepoRootEnv, Config.BuildTypeEnv });
+    
+        Output = output;
+        string buildPackagesDirectory = Path.Combine(Config.RepoRoot, "artifacts/source-build/self/src/artifacts/packages", Config.BuildType, "Shipping");
 
         if (!Directory.Exists(buildPackagesDirectory))
         {
-            throw new DirectoryNotFoundException($"Directory {buildPackagesDirectory} does not exist. Try building with './build.sh -sb' first.");
+            throw new DirectoryNotFoundException($"Directory {buildPackagesDirectory} does not exist, try building with './build.sh'.");
```
mthalman, line 39: This repo doesn't support building without the `-sb` switch.

tests/SbrpTests/ValidationTests.cs
```diff
@@ -18,71 +18,69 @@
...
+
+        var filteredPackages = Packages
+            .Where(package =>
             {
-                filteredPackages.Add(package);
-            }
-        }
+                string packageName = Path.GetFileNameWithoutExtension(package).ToLower();
+                packageName = Regex.Replace(packageName, VersionPattern, string.Empty);
+                return !targetPacks.Contains(packageName) && !textOnlyPackages.Contains(packageName);
```
mthalman, line 65: It would probably be more efficient to use a `HashSet` for `targetPacks` and `textOnlyPackages` instead of array enumeration since there is going to be so many lookups.

ellahathaway: You're right. Not sure why that using a HashSet didn't cross my mind.  Added in https://github.com/dotnet/source-build-reference-packages/pull/817/commits/8ab54ea0adfc0dd81813d7d749c06f7eb416dbce

tests/SbrpTests/ValidationTests.cs
ellahathaway: Fixed in https://github.com/dotnet/source-build-reference-packages/pull/817/commits/8ab54ea0adfc0dd81813d7d749c06f7eb416dbce

tests/SbrpTests/Utilities.cs
ellahathaway: Added in https://github.com/dotnet/source-build-reference-packages/pull/817/commits/8ab54ea0adfc0dd81813d7d749c06f7eb416dbce

tests/SbrpTests/DummyAttributeTypeProvider.cs
ellahathaway: That makes sense! I was under the assumption that it had to be `ICustomAttributeTypeProvider<Type>`. Changed to `ICustomAttributeTypeProvider<Type?>` and made related changes in https://github.com/dotnet/source-build-reference-packages/pull/817/commits/8ab54ea0adfc0dd81813d7d749c06f7eb416dbce.

ViktorHofer:
> Job 1: Build, run tests, and publish artifacts Windows

Note that SBRPs can't be built on Windows today as building the targeting pack is currently only supported on Unix. It should be able to change that though.

mthalman (approved):

ellahathaway:
> Note that SBRPs can't be built on Windows today as building the targeting pack is currently only supported on Unix. It should be possible to change that though.

This is mentioned in more detail in [GitHub Issue #3724](https://github.com/dotnet/source-build/issues/3724).

> The existing test project doesn't require the SBRP packages to be generated while the new test does require that. I would probably keep them separated.

I agree that even though the generate tests are independent of SBRP package generation, there is a case to be made for consolidating the test projects. As an example, the [SmokeTests in the VMR](https://github.com/dotnet/dotnet/tree/main/test/Microsoft.DotNet.SourceBuild.SmokeTests) are all contained within a single test project; this format promotes a more uniform code structure and reduces redundancy. Additionally, if there's a need to skip tests in the future, implementing tests as SkippableFacts/SkippableTheories would be a straightforward solution.

ViktorHofer:
> I agree that even though the generate tests are independent of SBRP package generation, there is a case to be made for consolidating the test projects. As an example, the [SmokeTests in the VMR](https://github.com/dotnet/dotnet/tree/main/test/Microsoft.DotNet.SourceBuild.SmokeTests) are all contained within a single test project; this format promotes a more uniform code structure and reduces redundancy. Additionally, if there's a need to skip tests in the future, implementing tests as SkippableFacts/SkippableTheories would be a straightforward solution.

We have hundreds of test projects in dotnet/runtime, because they test different components, i.e. `System.Text.RegularExpressions.Tests.csproj` tests the implementation of `System.Text.RegularExpressions.csproj`. The same is true for the generate-script tests and the tests that you are adding here. They are fundamentally different things.

> this format promotes a more uniform code structure and reduces redundancy

I have to disagree on that. What enables a uniform code structure are analyzer and .editorconfig rules, not a single test project.

> Additionally, if there's a need to skip tests in the future, implementing tests as SkippableFacts/SkippableTheories would be a straightforward solution.

ellahathaway:
@ViktorHofer

I've had a few discussions with @MichaelSimons and @mthalman on the points you bring up. I can understand the reasons for splitting these tests into two different projects; the generate tests are testing the generation logic, and the SBRP tests are testing the build logic. Since they are testing two different components of the repo, I agree that there is reason behind splitting up the tests into two different projects. However, this is a small repo, much smaller than dotnet/runtime. I think there because of this, there is value to the tests sharing a singular test project in this repo given its size.

> this format promotes a more uniform code structure and reduces redundancy

I recognize that this may not have come across as I had intended it to. I was not referring to the stylistic code structure of the tests, but rather the organization and grouping of data. My point was that by using a singular test project in this instance, we reduce the amount of repeated code, such as the RepoRoot variable and even the .proj files.

> Straightforward sure, but you want to skip tests based on parameters like the machine configuration / the runtime or other traits. You don't want to express your build dependencies and order via xunit traits.

It is still possible to skip tests based on these factors. For example, it's possible to determine the OS version and skip based on that.

Ultimately, I acknowledge there are reasons to both sides. If there is a significant preference for multiple test projects over a singular test project, I am happy to make that change. As of right now, I am choosing to leave the tests in a singular project.

ViktorHofer:
> Ultimately, I acknowledge there are reasons to both sides. If there is a significant preference for multiple test projects over a singular test project, I am happy to make that change. As of right now, I am choosing to leave the tests in a singular project.

I'm still not seeing how you can make this change without disabling the existing tests on Windows. See my question above:

> How would you keep the existing generate script tests enabled when merging the test projects? Disabling them on Windows, even temporarily isn't an improvement.

ellahathaway:
> How would you keep the existing generate script tests enabled when merging the test projects? Disabling them on Windows, even temporarily isn't an improvement.

That logic is [here](https://github.com/dotnet/source-build-reference-packages/pull/817/files#diff-ea39232432e80b1defd6c4b8e4465d3d6ea345fe095c606be42305c9276fc8bdR35.). The test is a skippable fact, skipped if the OS is Windows.

I also have pipeline runs with the reworked pipeline. The first pipeline run includes the logic of SBRP validation tests as they exist in this PR (Generate and validation tests). The second pipeline run runs the tests as they currently exist in main (just the Generate tests). 
- Pipeline run with SBRP validation tests: https://dev.azure.com/dnceng/internal/_build/results?buildId=2312957&view=results
- Pipeline run without SBRP validation tests: https://dev.azure.com/dnceng/internal/_build/results?buildId=2312971&view=results

ViktorHofer:
I see. So you chose to skip all the tests that aren't supported on Windows. I'm fine with that. My biggest concern was that you would need to disable the existing generate script tests on Windows.

> Ultimately, I acknowledge there are reasons to both sides. If there is a significant preference for multiple test projects over a singular test project, I am happy to make that change

That's the reason: > Since they are testing two different components of the repo

That said, I'm fine with your proposal given in its current form. I'm just not convinced that this is a good long term strategy. As an example, if I would want to add tests for [this msbuild task](https://github.com/dotnet/source-build-reference-packages/blob/main/src/packageSourceGenerator/PackageSourceGeneratorTask/PackageSourceGeneratorTask.csproj), I would not put them into this single test project as such tests would depend on msbuild and NuGet APIs. This is a different component.

> However, this is a small repo, much smaller than dotnet/runtime. I think there because of this, there is value to the tests sharing a singular test project in this repo given its size.

1. There is precedence, take a look at Arcade which is similar in size to this repo: https://github.com/dotnet/arcade/tree/main/src. Tests exist per component.
2. Repositories grow organically and I would be surprised if that won't be true for SBRP.

ellahathaway:
@ViktorHofer Thanks for your input; you bring up valid arguments for why there should be two test projects. However, after discussing this in more depth with @MichaelSimons, we've decided to stick with a single test project implementation as it exists in this PR given the current state of this repo.

tests/SbrpTests/ValidationTests.cs
```diff
@@ -0,0 +1,172 @@
...
+                throw new InvalidOperationException($"{SbrpAttributeType} is not formatted properly with a key, value pair.");
+            }
+        }
+
+        return false;
+    }
+
+    public async Task<bool> IsPackageSignedAsync(string packagePath, PackageSignatureVerifier verifier, SignedPackageVerifierSettings settings)
+    {
+        using var packageReader = new PackageArchiveReader(packagePath);
```
MichaelSimons, line 168: Consider making use of the language feature to simplify by removing the var usage.  This comment applies in several places.

```suggestion
        using PackageArchiveReader packageReader = new (packagePath);
```

tests/SbrpTests/ValidationTests.cs
```diff
@@ -0,0 +1,172 @@
...
+    public ValidationTests(ITestOutputHelper output) => Output = output;
+
+    [SkippableFact]
+    public async void ValidateSbrp()
+    {
+        Skip.If(RuntimeInformation.IsOSPlatform(OSPlatform.Windows), "This test is not supported on Windows.");
+        
+        InitializeValidationTests();
+
+        CheckForSbrpAttribute();
```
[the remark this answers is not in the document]
ellahathaway: I wanted to skip the whole test class if the test is running on windows. This means skipping the constructor and both tests in the class. But xunit does not have this feature. So I had two choices: either combine the constructor (now InitializeValidationTests) and the two tests into one test and skip that one test, or repeat the constructor’s work in each test and remove the shared members. I chose the first option because it seemed neater. However, I realize that this makes the test results less visible.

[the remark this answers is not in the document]
MichaelSimons: This solution doesn't scale well at least for now.  If we want to add additional pieces to validate, then it implies is all has to be in the same testcase.  We also may want additional validation that differs per package type.  Putting all logic in a single test case smells.

tests/SbrpTests/ValidationTests.cs
ellahathaway: Fixed in https://github.com/dotnet/source-build-reference-packages/pull/817/commits/2aae3ceba8c63acabf1af5acaeded0330785d706

tests/SbrpTests/ValidationTests.cs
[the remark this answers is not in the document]
ellahathaway: That makes sense. I've refactored into two test cases in https://github.com/dotnet/source-build-reference-packages/pull/817/commits/2aae3ceba8c63acabf1af5acaeded0330785d706

tests/SbrpTests/GenerateScriptTests.cs
```diff
@@ -28,13 +28,13 @@ public enum PackageType
...
     public string SandboxDirectory { get; set; }
-    public string RepoRoot { get; set; }
-    public ITestOutputHelper output { get; set; }
+    public ITestOutputHelper Output { get; set; }
 
     public GenerateScriptTests(ITestOutputHelper output)
     {
-        this.output = output;
-        RepoRoot = Environment.CurrentDirectory.Substring(0, Environment.CurrentDirectory.IndexOf("artifacts"));
+        Utilities.ValidateConfigParameters(new string[] { Config.RepoRootEnv });
```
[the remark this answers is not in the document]
ellahathaway: Removed the validation in Utilties.cs in https://github.com/dotnet/source-build-reference-packages/pull/817/commits/2aae3ceba8c63acabf1af5acaeded0330785d706 and let the tests themselves validate for null, empty, or whitespace.

tests/SbrpTests/SbrpTests.Tests.csproj
```diff
@@ -0,0 +1,13 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>$(NetCurrent)</TargetFramework>
+    <Nullable>enable</Nullable>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="NuGet.Packaging" Version="$(NuGetPackagingVersion)" />
+    <PackageReference Include="Xunit.SkippableFact" Version="1.4.13" />
```
ViktorHofer, line 10: Consider moving that property into Versions.props. That's what we do in other repos and that will make switching to NuGet Central Package Management in the future easier.

tests/SbrpTests/ValidationTests.cs
```diff
@@ -0,0 +1,178 @@
...
+{
+    private const string SbrpAttributeType = "System.Reflection.AssemblyMetadataAttribute";
+    private const string VersionPattern = @"(\.\d)+([\-\w])*";
+    public ITestOutputHelper Output { get; set; }
+
+    public ValidationTests(ITestOutputHelper output)
+    {
+        Output = output;
+        
+        if (string.IsNullOrWhiteSpace(Config.RepoRoot))
```
MichaelSimons, line 33: Ideally this logic wouldn't have to be duplicated across the tests.  Consider encapsulating it in the Config class.  Ideally getters wouldn't [throw](https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/property) but this isn't a framework so I think in this case in-lining the validation in the getter is acceptable.

ellahathaway: The main issue is that the environment variable could still be set to whitespace or empty, so it would not evaluate to the RHS of the null-coalescing operator in this instance.

I could do something such as:

```
public static string RepoRoot
    {
        get
        {
            string repoRoot = Environment.GetEnvironmentVariable(RepoRootEnv) ??
                Environment.CurrentDirectory.Substring(0, Environment.CurrentDirectory.IndexOf("artifacts"));
            if (string.IsNullOrWhiteSpace(repoRoot))
            {
                throw new InvalidOperationException($"Environment variable {RepoRootEnv} cannot be null, empty, or whitespace.");
            }
            return repoRoot;
        }
    }
```

In some of the vmr smoke tests, we follow the pattern of checking if the string is null, empty, or whitespace in the test itself.

ellahathaway: Alternatively, I could also rely on letting the test itself fail with a directory does not exist exception.

tests/SbrpTests/Config.cs
```diff
@@ -0,0 +1,33 @@
...
+
+using System;
+using System.IO;
+
+namespace SbrpTests;
+
+internal static class Config
+{
+    public const string RepoRootEnv = "SBRP_TESTS_REPO_ROOT";
+    public const string BuildTypeEnv = "SBRP_TESTS_BUILD_TYPE";
```
ViktorHofer, line 13: Who sets those environment variables? We in runtime use runtimeconfiguration.json values instead of env vars: https://github.com/dotnet/runtime/blob/a9791e69991ad542d6bbf1d9a8e77127210499bb/src/tools/illink/test/Mono.Linker.Tests/Mono.Linker.Tests.csproj#L44-L46

Those are then read like this: https://github.com/dotnet/runtime/blob/a9791e69991ad542d6bbf1d9a8e77127210499bb/src/tools/illink/test/Trimming.Tests.Shared/PathUtilities.cs#L29C1-L30

tests/SbrpTests/ValidationTests.cs
ellahathaway: Addressed in https://github.com/dotnet/source-build-reference-packages/pull/817/commits/417c9953c921b03eecb143ae7fb4dec5a5fc4fcc

tests/SbrpTests/SbrpTests.Tests.csproj
ellahathaway: Thanks for the comment, I've moved the property into Versions.props in https://github.com/dotnet/source-build-reference-packages/pull/817/commits/417c9953c921b03eecb143ae7fb4dec5a5fc4fcc

tests/SbrpTests/Config.cs
```diff
@@ -0,0 +1,33 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
```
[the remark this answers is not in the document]
ellahathaway: Thanks for pointing this out. I've opened an issue for adding this to SBRP: https://github.com/dotnet/source-build/issues/3759

I see in the .editorconfig you've provided that the header does not include "See the LICENSE file in the project root for more information.". However, this line exists in a substantial amount of places in the VMR (https://github.com/dotnet/dotnet/blob/16d04b93a08e02f95458f76e12c608757b67b36a/test/Microsoft.DotNet.SourceBuild.SmokeTests/Config.cs#L1 is an example).  Is there a standard across repos for what the header should be?

tests/SbrpTests/Config.cs
ellahathaway: They can be set locally by the user or they can be set by the pipeline (https://github.com/dotnet/installer/blob/main/eng/pipelines/templates/jobs/sdk-diff-tests.yml#L101 as an example of a pipeline setting environment variables).

tests/SbrpTests/Config.cs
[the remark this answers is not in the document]
ViktorHofer: Yes, the two-line license header is what repositories should use. See https://github.com/dotnet/runtime/pull/38793 for more details

tests/SbrpTests/Config.cs
```diff
@@ -0,0 +1,30 @@
...
+{
+    public const string RepoRootEnv = "SBRP_TESTS_REPO_ROOT";
+    public const string BuildTypeEnv = "SBRP_TESTS_BUILD_TYPE";
+
+    public static string RepoRoot
+    {
+        get
+        {
+            string repoRoot = Environment.GetEnvironmentVariable(RepoRootEnv) ??
+                Environment.CurrentDirectory.Substring(0, Environment.CurrentDirectory.IndexOf("artifacts"));
```
ViktorHofer, line 19: Consider removing this hardcoded path expectation and the environment variable in favor of runtimeconfiguration.json values. This path is known at build time but not at run time so a runtimeconfiguration value seems most appropriate.

ViktorHofer: Same for the BuildType which is also known at build time.

tests/SbrpTests/ValidationTests.cs
```diff
@@ -0,0 +1,178 @@
...
+        foreach (var package in packages)
+        {
+            bool isSigned = await IsPackageSignedAsync(package, verifier, settings);
+            Assert.False(isSigned, $"{package} is signed. Signed packages are not allowed in source-build-reference-packages.");
+        }
+    }
+
+    private string[] GetPackages()
+    {
+        string buildPackagesDirectory = Path.Combine(Config.RepoRoot, "artifacts/source-build/self/src/artifacts/packages", Config.BuildType, "Shipping");
```
ViktorHofer, line 110: I'm not a fan of hardcoding this path. This would break whenever we change the build layout (i.e. when removing the source build inner clone). We should use a runtimeconfiguration value for that instead so that it stays in sync with the msbuild build time property.

ellahathaway: I'm open to creating a new environment variable that defaults to this path. That said, it'd still have to be hardcoded somewhere. Given the scale of this repo, I'm comfortable leaving this as is.

tests/SbrpTests/Config.cs
ViktorHofer: All these paths are available at build time via msbuild properties and really shouldn't be hardcoded. It doesn't matter how small or large this repository is, we should never hardcode these paths.

- `Environment.CurrentDirectory.Substring(0, Environment.CurrentDirectory.IndexOf("artifacts"))` ->`$(RepoRoot)`
- `Path.Combine(Config.RepoRoot, "artifacts/source-build/self/src/artifacts/packages", Config.BuildType, "Shipping")` -> `$(ArtifactsShippingPackagesDir)`
- `BuildType` -> `$(Configuration)`

MichaelSimons: This nice thing about @ViktorHofer's suggestion is the runtime config properties can be set via msbuild logic in the csproj and read at runtime.  It would eliminate any hard coding of the default values to ensure correctness.  This feels like a better option given all these settings are known at build time.

tests/SbrpTests/ValidationTests.cs
MichaelSimons: Given this validation was moved to the config.cs file, is the check/throw necessary here anymore?

tests/SbrpTests/Config.cs
ellahathaway: I think the issue comes from having to work with the source-build inner clone in this case. 

Because the tests are run with "./test.sh", all of the MSBuild properties are set for the outer, non source-build repo. This is fine in the case of `RepoRoot` for the generate tests, but it gets more complicated when having to get the SB properties for the validation tests. For example, `ArtifactsShippingPackagesDir = /root/repos/source-build-reference-packages/artifacts/packages/Release/Shipping/`. I need it to be equal to `/root/repos/source-build-reference-packages/artifacts/source-build/self/src/artifacts/packages/Release/Shipping/`. This means that I'd still have to do some hardcoding to get the correct path.

I'm fine with using runtimeconfiguration.json values, but it would still involve doing some hardcoding.

ViktorHofer: > Because the tests are run with "./test.sh", all of the MSBuild properties are set for the outer, non source-build repo.

Can we change that so that the test invocation runs in the expected context (in this case, the source-build inner clone)?

ellahathaway: That gets a little tricky. Since the source-build build is run using arcade infra, I believe that arcade might have to support running tests as part of the inner clone build. At the same time, I don't know if there's a way to separate building the source-build inner clone from working in the context of the source-build inner clone. That is, I think you'd have to build the inner clone in order to work with it. We'd have to essentially change the infra of this test to run as a task that is triggered by a build arg such as `--validate` and try to run that task in the `InnerSourceBuildRepoRoot`

tests/SbrpTests/PathUtilities.cs
```diff
@@ -0,0 +1,26 @@
...
+    {
+        var artifactsDir = (string) AppContext.GetData ("SbrpTests.ArtifactsDir")!;
+        return Path.Combine (artifactsDir, "source-build", "self", "src");
+    }
+    
+    public static string GetSourceBuildPackagesShippingDir ()
+    {
+        var configuration = (string) AppContext.GetData ("SbrpTests.Configuration")!;
+        return Path.Combine (GetSourceBuildRepoRoot(), "artifacts", "packages", configuration, "Shipping");
+    }
```
mthalman, line 25: nit: remove the extra spaces
```suggestion
    public static string GetRepoRoot () =>
        (string)AppContext.GetData("SbrpTests.RepoRoot")!;

    public static string GetSourceBuildRepoRoot ()
    {
        var artifactsDir = (string)AppContext.GetData("SbrpTests.ArtifactsDir")!;
        return Path.Combine(artifactsDir, "source-build", "self", "src");
    }
    
    public static string GetSourceBuildPackagesShippingDir ()
    {
        var configuration = (string)AppContext.GetData("SbrpTests.Configuration")!;
        return Path.Combine(GetSourceBuildRepoRoot(), "artifacts", "packages", configuration, "Shipping");
    }
```

tests/SbrpTests/ValidationTests.cs
```diff
@@ -0,0 +1,166 @@
...
+            catch
+            {
+                throw new InvalidOperationException($"{SbrpAttributeType} is not formatted properly with a key, value pair.");
+            }
+        }
+
+        return false;
+    }
+
+    private async Task<bool> IsPackageSignedAsync(string packagePath, PackageSignatureVerifier verifier, SignedPackageVerifierSettings settings)
```
mthalman, line 160: Check for any other methods that can be `static` as well.

```suggestion
    private static async Task<bool> IsPackageSignedAsync(string packagePath, PackageSignatureVerifier verifier, SignedPackageVerifierSettings settings)
```

tests/SbrpTests/ValidationTests.cs
```diff
@@ -0,0 +1,166 @@
...
+    private const string SbrpRepoIdentifier = "source-build-reference-packages";
+    private const string VersionPattern = @"(\.\d)+([\-\w])*";
+    public ITestOutputHelper Output { get; set; }
+
+    public ValidationTests(ITestOutputHelper output) => Output = output;
+
+    [SkippableFact]
+    public void ValidateSbrpAttribute()
+    {
+        Skip.If(RuntimeInformation.IsOSPlatform(OSPlatform.Windows), "This test is not supported on Windows.");
```
mthalman, line 35: This is repeated for each test in this class. Would it work to place this just in the class constructor instead? In xUnit, the constructor gets invoked for each test if I remember correctly.

tests/SbrpTests/PathUtilities.cs
ellahathaway: I was following the style of https://github.com/dotnet/runtime/blob/a9791e69991ad542d6bbf1d9a8e77127210499bb/src/tools/illink/test/Trimming.Tests.Shared/PathUtilities.cs#L29C1-L30. Should I match how runtime formats this or should I remove the spaces?

mthalman: I've honestly never seen that style used before. It's not used throughout the runtime repo. My suggestion is to remove the spaces.

ViktorHofer: This runtime code originated from dotnet/linker. That style is weird, don't use it.

tests/SbrpTests/ValidationTests.cs
ellahathaway: I didn't know that it was possible to skip via the constructor, thanks for pointing this out!

ellahathaway:
This will fail until the arcade sdk from https://github.com/dotnet/installer/pull/17985 is merged in due to the change of the inner clone path. Until then, here is a success run with these changes: https://dev.azure.com/dnceng/internal/_build/results?buildId=2336530&view=results

mthalman (approved):

eng/Versions.props
```diff
@@ -21,5 +21,7 @@
     <MicrosoftNETCoreILDAsmVersion>6.0.0-preview.6.21352.12</MicrosoftNETCoreILDAsmVersion>
     <!-- SDK dependencies -->
     <MicrosoftDotNetGenAPITaskPackageVersion>9.0.100-alpha.1.23618.1</MicrosoftDotNetGenAPITaskPackageVersion>
+    <!-- xUnit dependencies -->
+    <XunitSkippableFactVersion>1.4.13</XunitSkippableFactVersion>
```
ViktorHofer, line 25: It's absolutely fine to use SkippableFact here. Just wanted to let you know that we already have in-built support for this and more in Arcade via [Microsoft.DotNet.XUnitExtensions](https://github.com/dotnet/arcade/tree/main/src/Microsoft.DotNet.XUnitExtensions).

tests/SbrpTests/SbrpTests.Tests.csproj
```diff
@@ -0,0 +1,28 @@
...
+    <Nullable>enable</Nullable>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="NuGet.Packaging" Version="$(NuGetPackagingVersion)" />
+    <PackageReference Include="Xunit.SkippableFact" Version="$(XunitSkippableFactVersion)" />
+  </ItemGroup>
+
+  <ItemGroup>
+
```
ViktorHofer, line 14: nit: remove empty line

```suggestion
```

ViktorHofer (approved):

eng/Versions.props
ellahathaway: This is great to know! Thanks for pointing this out.

tests/SbrpTests/SbrpTests.Tests.csproj
MichaelSimons, line 1: The file name feels a little odd to me.  Any reason it can be simplified to SbrpTests.csproj?

MichaelSimons (commented):
Looks good, One minor comment.

tests/SbrpTests/SbrpTests.Tests.csproj
ViktorHofer: The `.Tests` project name suffix is important in order for Arcade to recognize the project as a test project: https://github.com/dotnet/arcade/blob/3531486ae4b4e78149d815d99db3d3a0e8a3c64d/src/Microsoft.DotNet.Arcade.Sdk/tools/Tests.props#L16

 That said, `SBRP.Tests.csproj` would also work.

PR #822, opened 2023-11-08
Adding System.IO.Hashing

CyrusNajmabadi:


mthalman:
For tracking purposes, what is new package needed for?

mthalman (commented):
Per the instructions from https://github.com/dotnet/source-build-reference-packages#adding-new-packages:
* eng/Build.props needs to be updated to reference the new package version
* Modifications to existing package files need to be reverted

CyrusNajmabadi:
Roslyn wants to use System.IO.Hashing to use the xxhash128 type to do content hashing.  We currently use sha256, but would benefit greatly from a faster hash (xxhash is around 40x faster). We are not using this for cryptographic purposes. Thanks!

CyrusNajmabadi:
@mthalman does thsi look good now?

MichaelSimons:
The addition of a non-stable version raises alarms with me.  source-build-reference-packages should be stable released versions.  Is there a reason why roslyn doesn't utilize a dependency flow for this package?

CyrusNajmabadi:
>  Is there a reason why roslyn doesn't utilize a dependency flow for this package?

I don't know what "utilize a dependendency flow for this package" means.  Can you clarify?  

> The addition of a non-stable version raises alarms with me. 

The stable release (7.0) does not contain xxhash128, which is the library we are trying to move to in roslyn.

MichaelSimons:
> I don't know what "utilize a dependendency flow for this package" means. Can you clarify?

Declare the dependency in the versions.details.xml file adorned with the source-build metadata indicating it is required by source-build and optionally define a darc subscription to automatically update it.

[Example](https://github.com/dotnet/roslyn/blob/main/eng/Version.Details.xml#L21)

If this is not feasible/desirable, I would be amenable to accepting this PR as long as there is a tracking issue logged to update the dependency to the released 8.0.0 version once it GAs.

CyrusNajmabadi:
@MichaelSimons For context: i really don't know what any of these systems are doing, or the various pros/cons/reasonings surrounding any of this :)  I basically don't have any way to make an informed choice on what is best to do here.

For example "and optionally define a darc subscription".  I don't know what darc is.  I don't know what a darc subscription is :)

Basically, we simply want to reference this functionality so we can use it in roslyn, and that was blocked with a link saying we needed to update this repo. :)

> If this is not feasible/desirable,

I cannot make any determinations of feasibility or desirability.  I lack total context on being able to weigh in on that :)

> I would be amenable to accepting this PR as long as there is a tracking issue logged to update the dependency to the released 8.0.0 version once it GAs.

My end goal is simply being able to use the xxhash128 type, in a fashion that allows us to consume it on roslyn, while also meeting the needs of this repo.  But i'll need major assistance in getting there as I basically have no understanding of what the paths are within this repo, and i definitely cannot make determinations about which of those paths is the appropriate one to take :)

MichaelSimons:
Thinking about this I think the correct action is to accept this PR and log a follow up issue to upgrade the roslyn reference to the 8.0 GA version once it is released.  If the roslyn PR can wait, perhaps just waiting until Tuesday would be less work - your call but let us know so that we can merge this PR for you or if we should wait.

FYI to learn about dependency flow please see https://github.com/dotnet/arcade/blob/main/Documentation/DependencyFlowOnboarding.md.

CyrusNajmabadi:
We can wait until tuesday.  That's no issue for us.  :)

eng/Build.props
```diff
@@ -56,6 +56,8 @@
     <DependencyPackageProjects Include="$(RepoRoot)src\referencePackages\src\**\NuGet.DependencyResolver.Core.6.7.0.csproj" />
     <DependencyPackageProjects Include="$(RepoRoot)src\referencePackages\src\**\NuGet.ProjectModel.6.7.0.csproj" />
     <DependencyPackageProjects Include="$(RepoRoot)src\referencePackages\src\**\NuGet.Commands.6.7.0.csproj" />
+
+    <DependencyPackageProjects Include="$(RepoRoot)src\referencePackages\src\**\System.IO.Hashing.8.0.0-rc.2.23479.6.csproj" />
```
CyrusNajmabadi:
```suggestion
    <DependencyPackageProjects Include="$(RepoRoot)src\referencePackages\src\**\System.IO.Hashing.8.0.0.csproj" />
```

CyrusNajmabadi:
@mthalman @MichaelSimons this is updated to System.IO.Hashing 8.0.0.  Could you ptal?  Roslyn would very much like this.  Thanks! <3 :)

MichaelSimons (approved):

CyrusNajmabadi:
Thanks!

PR #824, opened 2023-11-10
Pipeline refactor

ellahathaway:
Closes https://github.com/dotnet/source-build/issues/3724

The goal of this PR is to migrate the tests for the source-build-reference-packages repo from their own individual jobs into steps in build jobs for Linux and Windows. Given that the repo does not currently build on Windows, that job remains as an individual test job. However, logic to integrate testing into a source build Linux job was accomplished in this PR.

The changes in this PR involve:
- Adding a stage to the source-build steps template to run and publish tests
- Adding logic to the source-build template to include a displayName
- Removing the generate-tests.yml and moving logic into ci.yml for Windows tests

Pipeline run with SBRP validation tests: https://dev.azure.com/dnceng/internal/_build/results?buildId=2312957&view=results
Pipeline run without SBRP validation tests: https://dev.azure.com/dnceng/internal/_build/results?buildId=2312971&view=results


azure-pipelines/builds/ci.yml
```diff
@@ -29,10 +29,32 @@ stages:
...
+    pool:
+      vmImage: $(imageName)
+    steps:
+    - checkout: self
+      clean: true
+    - script: $(Build.SourcesDirectory)/test.cmd
+      displayName: Run Tests
+    - task: PublishTestResults@2
+      displayName: Publish Test Results
+      inputs:
```
oleksandr-didyk, line 55: Nit: having `testRunTitle` here just like in the Linux job would probably improve the tests overview.
Currently it doesn't really tell you for what these test results are:
image

eng/common/templates/jobs/source-build.yml
```diff
@@ -21,6 +21,8 @@ parameters:
   # one job runs on 'defaultManagedPlatform'.
   platforms: []
 
+  displayName: 'Source-Build'
```
oleksandr-didyk, line 24: Nit: from my perspective `parameters` are inputs, i.e. something a user or a YAML calling in the template is providing. Static values are rather `variables`.

azure-pipelines/builds/ci.yml
```diff
@@ -29,10 +29,32 @@ stages:
...
+          agentOs: Windows_NT
+          imageName: "windows-latest"
+    pool:
+      vmImage: $(imageName)
+    steps:
+    - checkout: self
+      clean: true
+    - script: $(Build.SourcesDirectory)/test.cmd
+      displayName: Run Tests
+    - task: PublishTestResults@2
```
oleksandr-didyk, line 53: We don't want to run the publishing `always` like we do with the Linux job?

azure-pipelines/builds/ci.yml
```diff
@@ -29,10 +29,32 @@ stages:
...
+  ## Currently, the source-build-reference-packages repository does not support building on Windows, but it is capable of running the tests on Windows.
+  ## This job should be modified once the source-build-reference-packages repository supports building on Windows by transitioning to the template above.
+  - job: Test_Windows
+    strategy:
+      matrix:
+        Windows:
+          agentOs: Windows_NT
+          imageName: "windows-latest"
+    pool:
+      vmImage: $(imageName)
```
mthalman, line 47: Since this is running a single job we don't need a matrix strategy. This can be simplified:

```suggestion
    pool:
      vmImage: windows-latest
```

eng/common/templates/steps/source-build.yml
```diff
@@ -97,6 +97,16 @@ steps:
       /p:AssetManifestFileName=$assetManifestFileName
   displayName: Build
 
+- script: $(Build.SourcesDirectory)/test.sh
+  displayName: Run Tests
+- task: PublishTestResults@2
```
mthalman, line 102: The convention in this file is to have blank lines separating each step.

```suggestion

- task: PublishTestResults@2
```

eng/common/templates/jobs/source-build.yml
MichaelSimons: Jobs templates  unfortunately don't support variables so parameters is the only option aside from duplicating these static values.

eng/common/templates/steps/source-build.yml
```diff
@@ -97,6 +97,16 @@ steps:
       /p:AssetManifestFileName=$assetManifestFileName
   displayName: Build
 
+- script: $(Build.SourcesDirectory)/test.sh
```
MichaelSimons, line 100: [Changes are not allow](https://github.com/dotnet/source-build-reference-packages/tree/main/eng/common#readme) to the eng/common content.  It will get overridden by arcade dependency flows.

This is going to make this work significantly harder.

azure-pipelines/builds/ci.yml
```diff
@@ -29,10 +29,32 @@ stages:
...
+          logs: true
       enableSourceBuild: true
-  - template: /azure-pipelines/builds/generatescript-tests.yml
-    parameters:
-      buildConfiguration: Debug
+      sourceBuildParameters:
+        displayName: SourceBuild_Linux
+
+  ## The job below is a temporary workaround for running the tests on Windows.
+  ## Currently, the source-build-reference-packages repository does not support building on Windows, but it is capable of running the tests on Windows.
```
MichaelSimons, line 38:
```suggestion
  ## Currently, the source-build-reference-packages repository does not support building on Windows, but it is capable of running the generate tooling tests on Windows.
```

eng/common/templates/steps/source-build.yml
ellahathaway: Did not realize this... In that case, this might involve having to move away from the Arcade templates for the SB jobs. From my perspective, I did not see a way to add in the test step without modifying the arcade templates.

azure-pipelines/builds/ci.yml
```diff
@@ -29,10 +29,32 @@ stages:
...
+  ## This job should be modified once the source-build-reference-packages repository supports building on Windows by transitioning to the template above.
+  - job: Test_Windows
```
MichaelSimons, line 40: IMO, the legs should just be `Windows` and `Linux`.  The legs contents can differ.

eng/common/templates/steps/source-build.yml
MichaelSimons: It is plausible to make chances to the arcade templates in the arcade repo but they would have to be changes that would make sense for all repos.  For example, maybe it would make sense to provide a mechanism to invoke additional steps that are passed in.  It might not make sense to expose that throughout all of the templates but more so it you just wanted to reuses the steps template directly.

ellahathaway: This makes sense. I'll open an issue in arcade to explore some options.

azure-pipelines/builds/ci.yml
ellahathaway: Added in https://github.com/dotnet/source-build-reference-packages/pull/824/commits/b2930c5520b6e39c9e578184bbd8cef3186796db:

https://github.com/dotnet/source-build-reference-packages/blob/b2930c5520b6e39c9e578184bbd8cef3186796db/azure-pipelines/builds/ci.yml#L56

azure-pipelines/builds/ci.yml
ellahathaway: Added in https://github.com/dotnet/source-build-reference-packages/pull/824/commits/b2930c5520b6e39c9e578184bbd8cef3186796db:

https://github.com/dotnet/source-build-reference-packages/blob/b2930c5520b6e39c9e578184bbd8cef3186796db/azure-pipelines/builds/ci.yml#L55

azure-pipelines/builds/ci.yml
ellahathaway: Addressed in https://github.com/dotnet/source-build-reference-packages/pull/824/commits/b2930c5520b6e39c9e578184bbd8cef3186796db:

https://github.com/dotnet/source-build-reference-packages/blob/b2930c5520b6e39c9e578184bbd8cef3186796db/azure-pipelines/builds/ci.yml#L42

eng/common/templates/steps/source-build.yml
ellahathaway: Added in https://github.com/dotnet/source-build-reference-packages/pull/824/commits/b2930c5520b6e39c9e578184bbd8cef3186796db:

https://github.com/dotnet/source-build-reference-packages/blob/b2930c5520b6e39c9e578184bbd8cef3186796db/azure-pipelines/builds/ci.yml#L49

Left the steps template untouched due to https://github.com/dotnet/source-build-reference-packages/pull/824#discussion_r1391403747

azure-pipelines/builds/ci.yml
ellahathaway: Added in https://github.com/dotnet/source-build-reference-packages/pull/824/commits/b2930c5520b6e39c9e578184bbd8cef3186796db:

https://github.com/dotnet/source-build-reference-packages/blob/b2930c5520b6e39c9e578184bbd8cef3186796db/azure-pipelines/builds/ci.yml#L38

azure-pipelines/builds/ci.yml
ellahathaway: Addressed in https://github.com/dotnet/source-build-reference-packages/pull/824/commits/b2930c5520b6e39c9e578184bbd8cef3186796db:

https://github.com/dotnet/source-build-reference-packages/blob/b2930c5520b6e39c9e578184bbd8cef3186796db/azure-pipelines/builds/ci.yml#L35

https://github.com/dotnet/source-build-reference-packages/blob/b2930c5520b6e39c9e578184bbd8cef3186796db/azure-pipelines/builds/ci.yml#L40

eng/common/templates/steps/source-build.yml
MichaelSimons: FYI - https://github.com/dotnet/source-build/blob/main/Documentation/sourcebuild-in-repos/new-repo.md#advanced-granular-templates in case you haven't seen this documentation.

azure-pipelines/builds/ci.yml
```diff
@@ -19,20 +19,44 @@ stages:
...
-        publish:
-          artifacts: true
-          manifests: true
-      enableSourceBuild: true
-  - template: /azure-pipelines/builds/generatescript-tests.yml
-    parameters:
-      buildConfiguration: Debug
+  - job: Linux
+    pool:
+      vmImage: ubuntu-latest
```
MichaelSimons, line 24: I think it would be best to utilize the same docker image that arcade runs the sb legs in.  This will help provide isolation from changing dependencies.

ellahathaway: Addressed in https://github.com/dotnet/source-build-reference-packages/pull/824/commits/5f4f2f7d80af842110ae1fccd2145807c0d9f369

azure-pipelines/builds/ci.yml
```diff
@@ -15,24 +15,69 @@ variables:
...
-      publishAssetsImmediately: true
-      artifacts:
-        publish:
-          artifacts: true
-          manifests: true
-      enableSourceBuild: true
-  - template: /azure-pipelines/builds/generatescript-tests.yml
-    parameters:
-      buildConfiguration: Debug
+  - job: Linux
```
MichaelSimons, line 41: This job should be run inside the `mcr.microsoft.com/dotnet-buildtools/prereqs:centos-stream8` container to ensure the appropriate prereqs for source-build.  This is the container used by the source-build jobs template.

ellahathaway: Done in https://github.com/dotnet/source-build-reference-packages/pull/824/commits/6302848c6e0ff4f1d811c4835c0342103bd4b987. Ended up defining a container for windows as well.

MichaelSimons (approved):

azure-pipelines/builds/ci.yml
```diff
@@ -15,24 +15,69 @@ variables:
   - name: Codeql.Enabled
     value: true
 
+  - name: defaultPoolName
+    ${{ if eq(variables['System.TeamProject'], 'public') }}:
+      value: $[replace(replace(eq(contains(coalesce(variables['System.PullRequest.TargetBranch'], variables['Build.SourceBranch'], 'refs/heads/main'), 'release'), 'true'), True, 'NetCore-Svc-Public' ), False, 'NetCore-Public')]
+    ${{ if eq(variables['System.TeamProject'], 'internal') }}:
+      value: $[replace(replace(eq(contains(coalesce(variables['System.PullRequest.TargetBranch'], variables['Build.SourceBranch'], 'refs/heads/main'), 'release'), 'true'), True, 'NetCore1ESPool-Svc-Internal'), False, 'NetCore1ESPool-Internal')]
```
ViktorHofer, line 22: That is already handled by arcade: https://github.com/dotnet/arcade/blob/main/eng/common/templates/variables/pool-providers.yml

Include the variables file like this: https://github.com/dotnet/aspire/blob/74c40701ec6385bcccc2452863ffb3585d36601e/eng/pipelines/azure-pipelines.yml#L18

azure-pipelines/builds/ci.yml
```diff
@@ -15,24 +15,69 @@ variables:
...
+    ${{ if eq(variables['System.TeamProject'], 'public') }}:
+      value: ImageOverride -equals Build.Ubuntu.1804.Amd64.Open
+    ${{ if eq(variables['System.TeamProject'], 'internal') }}:
+      value: ImageOverride -equals Build.Ubuntu.1804.Amd64
+
+  - name: defaultPoolDemandsWindows
+    ${{ if eq(variables['System.TeamProject'], 'public') }}:
+      value: ImageOverride -equals 1es-windows-2022-open
+    ${{ if eq(variables['System.TeamProject'], 'internal') }}:
+      value: ImageOverride -equals 1es-windows-2022
```
ViktorHofer, line 34: These variables are only used in a single place. It would be less verbose to inline them and not define a separate variable for them. See https://github.com/dotnet/aspire/blob/74c40701ec6385bcccc2452863ffb3585d36601e/eng/pipelines/azure-pipelines.yml#L103-L108 for an example.

azure-pipelines/builds/ci.yml
```diff
@@ -15,24 +15,69 @@ variables:
...
+  - job: Linux
+    pool:
+      name: ${{ variables.defaultPoolName }}
+      demands: ${{ variables.defaultPoolDemandsLinux }}
+    container: mcr.microsoft.com/dotnet-buildtools/prereqs:centos-stream8
+    steps:
+    - template: /eng/common/templates/steps/source-build.yml
+
+    - script: $(Build.SourcesDirectory)/test.sh
+      displayName: Run Tests
```
ViktorHofer, line 49: Interesting. I would expect the `source-build.yml` file to support also running tests. It might make sense to add such a feature as an opt-in.

ViktorHofer: Same for the publish test results. The default `build` step supports running and publishing test results (and with different formats), i.e. here: https://github.com/dotnet/arcade/blob/69e73774ddfdacb1eea03e86d0c6ffcdb94eb4dd/eng/common/templates/job/job.yml#L32

cc @MichaelSimons @mthalman

ViktorHofer: Would it make more sense to add non source build legs here that then also run the tests and keep the existing source build legs (which wouldn't run tests) separate?

azure-pipelines/builds/ci.yml
```diff
@@ -15,24 +15,69 @@ variables:
...
-          manifests: true
-      enableSourceBuild: true
-  - template: /azure-pipelines/builds/generatescript-tests.yml
-    parameters:
-      buildConfiguration: Debug
+  - job: Linux
+    pool:
+      name: ${{ variables.defaultPoolName }}
+      demands: ${{ variables.defaultPoolDemandsLinux }}
+    container: mcr.microsoft.com/dotnet-buildtools/prereqs:centos-stream8
```
ViktorHofer, line 44: nit: we use mariner in other pipeline, i.e. aspire: https://github.com/dotnet/aspire/blob/74c40701ec6385bcccc2452863ffb3585d36601e/eng/pipelines/azure-pipelines.yml#L63

azure-pipelines/builds/ci.yml
MichaelSimons: I don't see value in running a non source-build leg for this repo since it's intended purpose is solely for source-build.  I do agree it would make sense to have an option to support running tests in the source-build leg.

azure-pipelines/builds/ci.yml
ellahathaway: (I am unable to reopen this merged PR. I have linked a [new PR with these changes](https://github.com/dotnet/source-build-reference-packages/pull/830/files#diff-33ae5736679b1efe12fde9f3fb58dbe360dd8a8bb32e5b98620bf7d644114f19R15).)

Good to know!

The reason I defined the values in this manner is because [`eng/common/templates/job/source-build.yml`](https://github.com/dotnet/source-build-reference-packages/blob/main/eng/common/templates/job/source-build.yml#L50) defines the values in this same way.

I'm surprised that `eng/common/templates/job/source-build.yml` doesn't also make use of these variables.

azure-pipelines/builds/ci.yml
ellahathaway: (I am unable to reopen this merged PR. I have linked a [new PR with these changes](https://github.com/dotnet/source-build-reference-packages/pull/830/files#diff-33ae5736679b1efe12fde9f3fb58dbe360dd8a8bb32e5b98620bf7d644114f19R26).)

This makes much more sense. Thank you for pointing this out. I have addressed this comment here: https://github.com/dotnet/source-build-reference-packages/pull/830/files#diff-33ae5736679b1efe12fde9f3fb58dbe360dd8a8bb32e5b98620bf7d644114f19R26

azure-pipelines/builds/ci.yml
ellahathaway: > It might make sense to add such a feature as an opt-in.

@ViktorHofer I opened https://github.com/dotnet/arcade/issues/14220 which relates to this suggestion.

azure-pipelines/builds/ci.yml
ellahathaway: https://github.com/dotnet/arcade/blob/69e73774ddfdacb1eea03e86d0c6ffcdb94eb4dd/eng/common/templates/jobs/source-build.yml#L17

The source-build template uses the centos-stream8 container, which is why I decided to also use it here.

azure-pipelines/builds/ci.yml
MichaelSimons: The source-build template existed prior to the pool-providers.yml.  It wasn't refactored to make use of it once it was defined.

PR #858, opened 2024-01-03
Add Microsoft.CodeAnalysis.Common.4.6.0 and Microsoft.CodeAnalysis.CSharp.4.6.0

ellahathaway:
Eliminates .NET 9 arcade prebuilts: https://github.com/dotnet/source-build/issues/3665


ellahathaway:
The reason `Microsoft.CodeAnalysis.CSharp` is failing is because it is referencing the public feed rather than the internal package-cache. https://github.com/dotnet/source-build-reference-packages/pull/859 fixes the issue.

mthalman (approved):

PR #859, opened 2024-01-05
Fix path for nupkg-cache, remove newtonsoft-json prebuilt

ellahathaway:
The path for `source-build-int-nupkg-cache` was previously being set to `/root/repos/source-build-reference-packages/artifacts/sb/src/artifacts/sb/src/artifacts/obj/source-built-upstream-cache/`, causing new SBRPs that depended on DependencyPackageProjectss to reference the online feeds.

This PR changes the path to be set to `/root/repos/source-build-reference-packages/artifacts/sb/src/artifacts/obj/source-built-upstream-cache/`. This fix should also eliminate the need for `Newtonsoft.Json.13.0.1` in the prebuilt baseline.

**Leaving as a draft because `Newtonsoft.Json.13.0.1` is still being detected as a prebuilt.**

### DIFF
--- a/eng/SourceBuild.props
+++ b/eng/SourceBuild.props
@@ -18,7 +18,7 @@
              Directories="$(LocalNuGetPackageCacheDirectory)" />
 
     <Exec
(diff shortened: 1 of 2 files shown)


ellahathaway:
Newtonsoft issue is likely caused by the previous versions of the packages in the PSB artifacts. Closing this PR and moving path changes to https://github.com/dotnet/source-build-reference-packages/pull/858

PR #861, opened 2024-01-08
Add .NET8 BCLs used by Roslyn

Cosifne:


Cosifne:
I still get failures like
 ##[error]artifacts/sb/src/src/referencePackages/src/system.text.json/8.0.0/lib/netstandard2.0/System.Text.Json.cs(1750,63): error CS0518: Predefined type 'System.Runtime.CompilerServices.IsExternalInit' is not defined or imported [/__w/1/s/artifacts/sb/src/src/referencePackages/src/system.text.json/8.0.0/System.Text.Json.8.0.0.csproj::TargetFramework=netstandard2.0]

Does anyone know how to address this?

JoeRobich:
> Does anyone know how to address this?

Looks like they manually change `init` to `set { }` when adding previous versions. See https://github.com/dotnet/source-build-reference-packages/pull/751/commits/0bf863632e47a2d59824e3cc39d1544a837871dd

src/referencePackages/src/system.runtime/4.3.0/ref/netstandard1.0/System.Runtime.cs
```diff
@@ -4,9 +4,6 @@
 // ------------------------------------------------------------------------------
 // Changes to this file must follow the http://aka.ms/api-review process.
 // ------------------------------------------------------------------------------
-using System;
-using System.Reflection;
-using System.Runtime.CompilerServices;
 [assembly: System.Runtime.CompilerServices.ReferenceAssembly]
```
JoeRobich, line 7: Likely need to revert changes to these system.runtime files.

src/referencePackages/src/system.memory/4.5.5/lib/netstandard1.1/System.Memory.cs
```diff
@@ -303,7 +303,7 @@ public static bool StartsWith<T>(this Span<T> span, ReadOnlySpan<T> value)
         public readonly bool TryCopyTo(Memory<T> destination) { throw null; }
     }
 
-    public readonly ref partial struct ReadOnlySpan<T>
+    public readonly partial struct ReadOnlySpan<T>
```
JoeRobich, line 306: May need to revert these system.memory changes as well

src/referencePackages/src/system.runtime.interopservices/4.3.0/ref/netcoreapp1.1/System.Runtime.InteropServices.cs
```diff
@@ -103,6 +103,16 @@ public enum CallingConvention
         ThisCall = 4
     }
 
+    [AttributeUsage(AttributeTargets.Assembly | AttributeTargets.Class, Inherited = false)]
+    public sealed partial class ClassInterfaceAttribute : Attribute
```
JoeRobich, line 107: And possibly revert these system.runtime.interopservice changes

Cosifne:
-> Thanks for Joey to find out how to make this things work.

src/referencePackages/src/system.text.json/8.0.0/lib/netstandard2.0/System.Text.Json.cs
```diff
@@ -1747,15 +1747,15 @@ public partial interface IJsonTypeInfoResolver
 
     public sealed partial class JsonCollectionInfoValues<TCollection>
     {
-        public JsonTypeInfo ElementInfo { get { throw null; } init { } }
+        public JsonTypeInfo ElementInfo { get { throw null; } set { } }
```
ViktorHofer, line 1750: Is this a GenAPI bug? cc @ericstj @andriipatsula

ViktorHofer:
Roslyn decided to stay on the `.NET 8 Eng` channel instead of `.NET Eng - Latest` so you will need to backport this into the 8.0 branch of SBRP after merging.

mthalman:
> Roslyn decided to stay on the `.NET 8 Eng` channel instead of `.NET Eng - Latest` so you will need to backport this into the 8.0 branch of SBRP after merging.

This can't be backported to the 8.0 branch because it relies on the .NET 8 targeting pack which is only defined in the main branch. I'm not sure SBRP is even the right solution here. Roslyn is in such a weird state. Roslyn's VS 17.9 channel is flowing to both the release/8.0.2xx and main branches of sdk. Just looking at it from a 8.0.2xx perspective, it would seem to make sense that these new dependencies in https://github.com/dotnet/roslyn/pull/71516 should just be dependency flow from the other repos. I don't know how that works in a post-release servicing world. And then you've got other things like https://github.com/dotnet/roslyn/pull/71516/files#diff-fb62e94a1d6f29f863e3d0a22aa38269f6cd1d7f03b109dc06e2cbf2548b86d3L9 which has this: `<Dependency Name="Microsoft.SourceBuild.Intermediate.source-build-reference-packages" Version="9.0.0-alpha.1.23627.2">`, and that doesn't make sense when targeting 8.0.

src/referencePackages/src/system.text.json/8.0.0/lib/netstandard2.0/System.Text.Json.cs
ericstj: That should have been fixed long ago : https://github.com/dotnet/roslyn/pull/68960

I checked with the latest GenAPI commandline tool and was able to get it to emit the correct `init` accessor for both the netstandard2.0 build and the net8.0 build of STJ.

Cosifne:
Could anyone point me what would be the right direction to make this in?

mthalman:
@MichaelSimons - Do you have past knowledge on how things were handled with SBRP and Roslyn when we reach servicing, particularly when their channel flows to two different SDK branches?

jaredpar:
> Roslyn is in such a weird state. Roslyn's VS 17.9 channel is flowing to both the release/8.0.2xx and main branches of sdk

This is our normal state that we operate on virtually 100% of the time.

PR #866, opened 2024-01-12
Add System.Reflection.MetadataLoadContext 8.0.0 and deps

kg:
WasmAppBuilder in dotnet/runtime needs S.R.MLC 8.0.0 (https://github.com/dotnet/runtime/pull/94446) so it needs to be added to source build reference packages.

Not sure I did this correctly, because build.sh doesn't work in my linux environment (4300 build failures).

### DIFF
--- a/src/referencePackages/src/system.memory/4.5.5/lib/netstandard1.1/System.Memory.cs
+++ b/src/referencePackages/src/system.memory/4.5.5/lib/netstandard1.1/System.Memory.cs
@@ -303,7 +303,7 @@ namespace System
         public readonly bool TryCopyTo(Memory<T> destination) { throw null; }
     }
 
-    public readonly partial struct ReadOnlySpan<T>
+    public readonly ref partial struct ReadOnlySpan<T>
     {
         private readonly object _dummy;
         private readonly int _dummyPrimitive;
@@ -352,7 +352,7 @@ namespace System
 
         public readonly bool TryCopyTo(Span<T> destination) { throw null; }
 
-        public partial struct Enumerator
+        public ref partial struct Enumerator
         {
             private ReadOnlySpan<T> _span;
             private object _dummy;
@@ -380,7 +380,7 @@ namespace System
         public readonly object GetObject() { throw null; }
     }
 
-    public readonly partial struct Span<T>
+    public readonly ref partial struct Span<T>
     {
         private readonly object _dummy;
         private readonly int _dummyPrimitive;
@@ -435,7 +435,7 @@ namespace System
 
         public readonly bool TryCopyTo(Span<T> destination) { throw null; }
 
-        public partial struct Enumerator
+        public ref partial struct Enumerator
         {
             private Span<T> _span;
             private object _dummy;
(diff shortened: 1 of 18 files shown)


src/referencePackages/src/system.memory/4.5.5/lib/netstandard1.1/System.Memory.cs
```diff
@@ -303,7 +303,7 @@ public static bool StartsWith<T>(this Span<T> span, ReadOnlySpan<T> value)
         public readonly bool TryCopyTo(Memory<T> destination) { throw null; }
     }
 
-    public readonly ref partial struct ReadOnlySpan<T>
+    public readonly partial struct ReadOnlySpan<T>
```
kg, line 306: I'm not sure why the generator did this.

kg:
The generator seems very broken here. Is there something wrong with running it on Debian? Do I need to install a specific set of dependencies for it to generate code correctly?

src/referencePackages/src/system.runtime/4.3.0/ref/netstandard1.0/System.Runtime.cs
```diff
@@ -4,9 +4,6 @@
 // ------------------------------------------------------------------------------
 // Changes to this file must follow the http://aka.ms/api-review process.
 // ------------------------------------------------------------------------------
-using System;
```
kg, line 7: It seems wrong that the generator is doing this.

MichaelSimons:
> > Not sure I did this correctly, because build.sh doesn't work in my linux environment (4300 build failures).
> 
> @kg please share a binlog so that we can diagnose those failures.

Let's use https://github.com/dotnet/source-build/issues/3969 to discuss the local build failures.

mthalman:
> The changes in System.Runtime.InteropServices should be reverted. Unsure why those changed.

Why should they not be defined?

mthalman:
> > Why should they not be defined?
> 
> These changes are applied to files that correspond to a package that last shipped in 2016. I don't think changes to that package should be necessary for adding S.R.MetadataLoadContext/8.0.0 which shipped in 2023. Or am I missing something?

According to the [readme](https://github.com/dotnet/source-build-reference-packages?tab=readme-ov-file#workflow), they should be updated. It's better to keep the files up-to-date wrt the generator when possible to avoid having to keep deal with these diffs.

ViktorHofer:
Thanks for the explanation. This really is point number 2 from the doc:

> The generate tooling has changed since the last time this package was generated. The new changes should be considered better/correct and should be committed.

The changes are due to an attribute fix in GenAPI that I made late in .NET 8. The attributes were missing so this changes is good.

MichaelSimons (approved):

MichaelSimons:
@mthalman, @ViktorHofer - sounds like you both are in agreement that the changes look good and we have tracking issues for the related items mentioned here - build not working on debian and automation to regenerate the packages on a regular cadence.  If you would could you give a thumbs up or approve the PR and so we can merge this PR to unblock the dependent repo work.  TIA

mthalman (approved):

kg:
Just to verify: I would now create a dotnet/runtime PR to update its SBRP reference to the merge commit aefcb948d0ba9e0215f5081004f94309873124fc ?

mthalman:
Dependency flow will do this automatically. Looks like the darc subscription from source-build-reference-packages to runtime is set to only update once a week. If you want it sooner, you can manually trigger the subscription [once everything gets published](https://dev.azure.com/dnceng/internal/_build/results?buildId=2354973&view=results) (internal link). You'd need to do that using the [darc tool](https://github.com/dotnet/arcade/blob/main/Documentation/Darc.md).

PR #873, opened 2024-01-22
Manually add API that isn't present in the contract assembly

ViktorHofer:
... but that is required by the compiler (base type abstract member). For more details see discussion in
https://github.com/dotnet/source-build/issues/4000

Fixes https://github.com/dotnet/source-build/issues/4000

Without this change, GenAPI removes these API members every time the package is re-generated.

### DIFF
--- a/src/referencePackages/src/system.runtime.interopservices/4.3.0/System.Runtime.InteropServices.Manual.cs
+++ b/src/referencePackages/src/system.runtime.interopservices/4.3.0/System.Runtime.InteropServices.Manual.cs
@@ -1,0 +1,20 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+// ------------------------------------------------------------------------------
+// Changes to this file must follow the http://aka.ms/api-review process.
+// ------------------------------------------------------------------------------
+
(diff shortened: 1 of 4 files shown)


src/referencePackages/src/system.runtime.interopservices/4.3.0/System.Runtime.InteropServices.Manual.cs
```diff
@@ -0,0 +1,20 @@
...
+// See the LICENSE file in the project root for more information.
+// ------------------------------------------------------------------------------
+// Changes to this file must follow the http://aka.ms/api-review process.
+// ------------------------------------------------------------------------------
+
+// See https://github.com/dotnet/source-build/issues/4000 for why this is necessary.
+
+namespace System.Runtime.InteropServices
+{
+    public partial class ComAwareEventInfo : Reflection.EventInfo
```
mthalman, line 12: Can you include a comment to clean this up when https://github.com/dotnet/source-build/issues/4000 is fixed?

ViktorHofer: This PR closes https://github.com/dotnet/source-build/issues/4000. We need to keep this as long as we build System.Runtime.InteropServices because it incorrectly shipped in .NET Core 1.1 days. This isn't a GenAPI bug.

mthalman (approved):

PR #876, opened 2024-01-25
Pin targeting pack version for net8.0

mthalman:
Fixes https://github.com/dotnet/source-build/issues/4037

I'm running a test build now to verify this does indeed fix the issue.


mthalman:
This does not seem to fix the issue and I'm not sure why. Here's my [test build](https://dev.azure.com/dnceng/internal/_build/results?buildId=2362133&view=results) (internal link). It's still showing up as a prebuilt in the stage 2 build.

@MichaelSimons - Do you have any ideas on what's going on?

MichaelSimons:
IIRC, this change will only affects SBRP.  Do you know what repo is bring in 8.0.1?  Building locally an looking at the repo level prebuilt reports is a way to narrow it down.  We don't capture these reports in CI artifacts.

mthalman:
This does resolve the SBRP issue. There's another similar issue to fix in SBE.

MichaelSimons (approved):

PR #890, opened 2024-02-12
Add missing nuget 6.8.0 packages

akoeplinger:
https://github.com/dotnet/source-build-reference-packages/pull/831 added most 6.8.0 packages but it missed a couple.
This adds them.

### DIFF
--- a/src/referencePackages/src/nuget.commands/6.8.0/NuGet.Commands.6.8.0.csproj
+++ b/src/referencePackages/src/nuget.commands/6.8.0/NuGet.Commands.6.8.0.csproj
@@ -1,0 +1,23 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFrameworks>net5.0;netstandard2.0</TargetFrameworks>
+    <AssemblyName>NuGet.Commands</AssemblyName>
+    <StrongNameKeyId>MicrosoftShared</StrongNameKeyId>
+  </PropertyGroup>
+
+  <ItemGroup Condition="'$(TargetFramework)' == 'net5.0'">
+    <PackageReference Include="NuGet.Credentials" Version="6.8.0" />
+    <PackageReference Include="NuGet.ProjectModel" Version="6.8.0" />
+    <PackageReference Include="Microsoft.Extensions.FileProviders.Abstractions" Version="6.0.0" />
+    <PackageReference Include="Microsoft.Extensions.FileSystemGlobbing" Version="6.0.0" />
+  </ItemGroup>
+
+  <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
+    <PackageReference Include="NuGet.Credentials" Version="6.8.0" />
+    <PackageReference Include="NuGet.ProjectModel" Version="6.8.0" />
+    <PackageReference Include="Microsoft.Extensions.FileProviders.Abstractions" Version="6.0.0" />
+    <PackageReference Include="Microsoft.Extensions.FileSystemGlobbing" Version="6.0.0" />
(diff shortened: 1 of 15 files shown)


akoeplinger:
@mthalman I don't understand the prebuilt failure about the new package in the `source-build-reference-packages-CI (Build Linux)` leg, mind taking a look?

MichaelSimons:
> @mthalman I don't understand the prebuilt failure about the new packages in the `source-build-reference-packages-CI (Build Linux)` leg, mind taking a look?

Try this per the [Adding new Packages](https://github.com/dotnet/source-build-reference-packages#adding-new-packages) instructions

```
Add DependencyPackageProjects for all new projects in the [eng/Build.props](https://github.com/dotnet/source-build-reference-packages/blob/main/eng/Build.props#L9) in the correct dependency order.
```

akoeplinger:
@MichaelSimons hmm none of these packages are new though, just newer versions, and I don't see the existing projects in Build.props

akoeplinger:
Ah I understand now, the references can be cleaned up once we're re-bootstrapping, that's why I couldn't find the existing ones: https://github.com/dotnet/source-build-reference-packages/commit/f2c3fed62861b918dfe300f01b497551813a56df

MichaelSimons:
> @MichaelSimons hmm none of these packages are new though, just newer versions, 

A new package == new nupkg in this context.

> and I don't see the existing projects in Build.props

This is only contains new projects added that aren't already in the previous source-built artifacts.  This list gets trimmed as part of the SB release process when the previous source-built artifacts are updated.

This whole pattern is not ideal and we have backlog items to address this but it is non-trivial.

akoeplinger:
Cool, it's green now, thanks for your help!

eng/Build.props
```diff
@@ -25,6 +25,11 @@
     <DependencyPackageProjects Include="$(RepoRoot)src\referencePackages\src\**\Microsoft.CodeAnalysis.Common.4.6.0.csproj" />
     <DependencyPackageProjects Include="$(RepoRoot)src\referencePackages\src\**\Microsoft.CodeAnalysis.CSharp.4.6.0.csproj" />
 
+    <DependencyPackageProjects Include="$(RepoRoot)src\referencePackages\src\**\NuGet.Commands.6.8.0.csproj" />
+    <DependencyPackageProjects Include="$(RepoRoot)src\referencePackages\src\**\NuGet.LibraryModel.6.8.0.csproj" />
+    <DependencyPackageProjects Include="$(RepoRoot)src\referencePackages\src\**\NuGet.DependencyResolver.Core.6.8.0.csproj" />
+    <DependencyPackageProjects Include="$(RepoRoot)src\referencePackages\src\**\NuGet.ProjectModel.6.8.0.csproj" />
```
mthalman, line 31: These need to be in dependency order:

```suggestion
    <DependencyPackageProjects Include="$(RepoRoot)src\referencePackages\src\**\NuGet.LibraryModel.6.8.0.csproj" />
    <DependencyPackageProjects Include="$(RepoRoot)src\referencePackages\src\**\NuGet.DependencyResolver.Core.6.8.0.csproj" />
    <DependencyPackageProjects Include="$(RepoRoot)src\referencePackages\src\**\NuGet.ProjectModel.6.8.0.csproj" />
    <DependencyPackageProjects Include="$(RepoRoot)src\referencePackages\src\**\NuGet.Commands.6.8.0.csproj" />
```

akoeplinger: Thanks, I tried to do that but missed the NuGet.Commands -> NuGet.ProjectModel dep.

akoeplinger:
Looks like I don't have merge permissions here :)

PR #879, opened 2024-01-30
Add System.Drawing.Common and M.W32.SystemEvents 8.0.0

ViktorHofer:
Needed for https://github.com/dotnet/sdk/pull/38348

PR #892, opened 2024-02-13
Add System.Drawing.Common 8.0.1

jjonescz:
- Needed for https://github.com/dotnet/sdk/pull/38156#issuecomment-1940806649
- Similar to https://github.com/dotnet/source-build-reference-packages/pull/879

cc @ViktorHofer

### DIFF
--- a/eng/Build.props
+++ b/eng/Build.props
@@ -24,6 +24,7 @@
       Format:
       <DependencyPackageProjects Include="$(RepoRoot)src\referencePackages\src\**\Microsoft.Extensions.Options.5.0.0.csproj" />
     -->
+    <DependencyPackageProjects Include="$(RepoRoot)src\referencePackages\src\**\Microsoft.Win32.SystemEvents.8.0.0.csproj" />
   </ItemGroup>
 
   <ItemGroup Condition="'$(BuildDependencyPackageProjects)' == 'true'">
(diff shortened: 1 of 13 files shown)


eng/Build.props
```diff
@@ -24,6 +24,7 @@
       Format:
       <DependencyPackageProjects Include="$(RepoRoot)src\referencePackages\src\**\Microsoft.Extensions.Options.5.0.0.csproj" />
     -->
+    <DependencyPackageProjects Include="$(RepoRoot)src\referencePackages\src\**\Microsoft.Win32.SystemEvents.8.0.0.csproj" />
```
ViktorHofer, line 27: You need to add System.Drawing.Common here as well. It needs to be defined SystemEvents.

mthalman (commented):
I don't see a need to add this in order to satisfy the prebuilt in https://github.com/dotnet/sdk/pull/38156 since that PR is targeting 8.0.3xx. For .NET 8, the VMR only supports 8.0.1xx. The general rule for SBRP is to only add content here that is consumed by the VMR. So instead I think it would be better to just mark this as an [allowed prebuilt](https://github.com/dotnet/sdk/blob/main/eng/SourceBuildPrebuiltBaseline.xml) in the sdk repo.

ViktorHofer:
@jjonescz sorry for the unnecessary extra work. I wasn't aware of that policy.

PR #899, opened 2024-02-22
Remove unused NoTargets and Traversal packages

MichaelSimons:
Fixes https://github.com/dotnet/source-build/issues/4085


MichaelSimons:
Verified in https://github.com/dotnet/installer/pull/18755.

NikolaMilosavljevic:
Was this also verified in a repo-specific build of a repo that consumes older version of these packages? For instance, `diagnostics` and `cecil` both use 3.5.0 of `NoTargets` SDK. Would it cause a prebuilt failure in source-build leg?

NikolaMilosavljevic (approved):

MichaelSimons:
> Was this also verified in a repo-specific build of a repo that consumes older version of these packages? For instance, `diagnostics` and `cecil` both use 3.5.0 of `NoTargets` SDK. Would it cause a prebuilt failure in source-build leg? I guess prebuilt failure can be fixed at the time of SBRP flow into affected repo.

Those repos should upgrade when they pick up this version of SBRP.  SBRP should not contain package that are only used in repo level source-build legs.

PR #905, opened 2024-02-27
Rename SourceBuild.props to DotNetBuild.props

MichaelSimons:
Related to https://github.com/dotnet/source-build/issues/3963


eng/DotNetBuild.props
mthalman, line 1: Shouldn't this be SourceOnlyBuild.props since this repo is specific to source build?

MichaelSimons: No, this file is being renamed to be more general purposed as part of the unified build work.  This file is an [arcade construct](https://github.com/dotnet/arcade/blob/main/src/Microsoft.DotNet.Arcade.Sdk/tools/SourceBuild/SourceBuildArcade.targets#L18) - e.g. the name is fixed.  There is no need/value in having different arcade constructs for source-only repos.

PR #957, opened 2024-05-14
Update CODEOWNERS

MichaelSimons:


mthalman (commented):
Why not source-build-internal?

MichaelSimons:
> Why not source-build-internal?

I updated to use it.  Originally it was not used because if one reviews, the PR is removed from everyone else's notification list.

mthalman (approved):

PR #990, opened 2024-07-22
[8.0] Add System.Formats.Asn1.6.0.1

ellahathaway:
Related to https://github.com/dotnet/source-build/issues/4518

### DIFF
--- a/src/referencePackages/src/system.security.cryptography.pkcs/6.0.1/System.Security.Cryptography.Pkcs.6.0.1.csproj
+++ b/src/referencePackages/src/system.security.cryptography.pkcs/6.0.1/System.Security.Cryptography.Pkcs.6.0.1.csproj
@@ -7,18 +7,18 @@
   </PropertyGroup>
 
   <ItemGroup Condition="'$(TargetFramework)' == 'net6.0'">
-    <PackageReference Include="System.Formats.Asn1" Version="6.0.0" />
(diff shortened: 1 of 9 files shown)


MichaelSimons (approved):

mthalman (approved):

src/referencePackages/src/system.security.cryptography.pkcs/6.0.1/System.Security.Cryptography.Pkcs.6.0.1.csproj
```diff
@@ -7,18 +7,18 @@
   </PropertyGroup>
 
   <ItemGroup Condition="'$(TargetFramework)' == 'net6.0'">
-    <PackageReference Include="System.Formats.Asn1" Version="6.0.0" />
+    <PackageReference Include="System.Formats.Asn1" Version="6.0.1" />
```
MichaelSimons, line 10: We should be following the guidance at https://github.com/dotnet/source-build-reference-packages?tab=readme-ov-file#vulnerable-packages and documenting the CVE that prompted the upgrade.  The primary reason for this is so that we don't loose these customizations if/when the package is regenerated.

ellahathaway: Opened #992  and #993 to address

PR #1006, opened 2024-08-13
Update Nuget.Packaging reference version

ellahathaway:
Related to https://github.com/dotnet/source-build/issues/4518 (see https://github.com/dotnet/source-build/issues/4518#issuecomment-2287250159)

The alerts also extended to a transitive dependency for System.Formats.Asn1.6.0.0 which was being pulled in by the version of Nuget.Packaging we were using in our test project and the package source generator project.

I'm not sure if this version of Nuget.Packaging needs to be added as an SBRP or not.


mthalman (approved):

MichaelSimons:
> I'm not sure if this version of Nuget.Packaging needs to be added as an SBRP or not.

No, this is a "real" package reference used by the tests and generation tooling.

PR #1032, opened 2024-09-09
Add system.diagnostics.diagnosticsource/5.0.0

MichaelSimons:
system.diagnostics.diagnosticsource/5.0.0 was removed in https://github.com/dotnet/source-build-reference-packages/pull/1026 but is still required by the product (https://github.com/dotnet/sdk/pull/43224).

I will investigate how this is referenced and why it was reported as an unreferenced package.

This does not need to be added as a DependencyPackageProject because it is being re-added (e.g. it already exists in n-1)


MichaelSimons:
> I will investigate how this is referenced and why it was reported as an unreferenced package.

I see what I did wrong.  There is a difference in the unreferenced packages depending on if the build is using Microsoft built Sdk versus source built Sdk.  To correctly determine the unreference packages, both types of legs must be examined.

mthalman (approved):

PR #1038, opened 2024-09-17
Add Microsoft.NETCore.Platforms 3.1.0 text only package

MichaelSimons:
Related to https://github.com/dotnet/source-build/issues/4617.


mthalman (commented):
Does this need to be added to build.props?

MichaelSimons:
> Does this need to be added to build.props?

In the past we haven't been adding text-only packages unless they are referenced by ref packages.

mthalman (approved):

PR #1055, opened 2024-10-10
Add Microsoft.Private.Intellisense.9.0.0-preview-20241010.1

carlossanlop:
PTAL @dotnet/source-build 

This should be the last intellisense package update before we ship .NET 9.

Note: Last time I deleted the most recent existing folder and it caused some trouble when doing updates. I decided not to delete it this time, @MichaelSimons, so it doesn't cause issues.


MichaelSimons (approved):
Agreed that keeping the previous versions is less risky.  We can clean those up later.  I assume you will backport to 9.0?

PR #1060, opened 2024-10-14
Add STJ 8.0.5

marcpopMSFT:
I ran the generate command, added the STJ items, and updated the build.props.  We need it for this and various other PRs into the templating repo which has a hardcoded STJ version as it doesn't get it through codeflow: https://github.com/dotnet/templating/pull/8430

I'm a bit surprised no one else needed a the SBRP update for STJ.

### DIFF
--- a/src/referencePackages/src/system.text.json/8.0.5/lib/netstandard2.0/System.Text.Json.cs
+++ b/src/referencePackages/src/system.text.json/8.0.5/lib/netstandard2.0/System.Text.Json.cs
@@ -1747,15 +1747,15 @@ namespace System.Text.Json.Serialization.Metadata
 
     public sealed partial class JsonCollectionInfoValues<TCollection>
     {
-        public JsonTypeInfo ElementInfo { get { throw null; } init { } }
+        public JsonTypeInfo ElementInfo { get { throw null; } set { } }
 
-        public JsonTypeInfo? KeyInfo { get { throw null; } init { } }
+        public JsonTypeInfo? KeyInfo { get { throw null; } set { } }
 
-        public JsonNumberHandling NumberHandling { get { throw null; } init { } }
+        public JsonNumberHandling NumberHandling { get { throw null; } set { } }
 
-        public Func<TCollection>? ObjectCreator { get { throw null; } init { } }
+        public Func<TCollection>? ObjectCreator { get { throw null; } set { } }
 
-        public Action<Utf8JsonWriter, TCollection>? SerializeHandler { get { throw null; } init { } }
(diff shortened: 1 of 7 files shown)


src/referencePackages/src/system.text.json/8.0.5/lib/netstandard2.0/System.Text.Json.cs
```diff
@@ -0,0 +1,2095 @@
...
+    }
+
+    public partial interface IJsonTypeInfoResolver
+    {
+        JsonTypeInfo? GetTypeInfo(Type type, JsonSerializerOptions options);
+    }
+
+    public sealed partial class JsonCollectionInfoValues<TCollection>
+    {
+        public JsonTypeInfo ElementInfo { get { throw null; } init { } }
```
mthalman, line 1750: The `init` is incorrectly generated here and should be changed to `set`. See https://github.com/dotnet/source-build-reference-packages/pull/861#issuecomment-1882031209.

marcpopMSFT: Is there a plan to update the tooling to handle this? This is already far too much friction for updating a single version number and it's only going to get worse as the push to get these updated asap ramps up.

marcpopMSFT: Also, the generate command when I ran it generated a bunch of additional files that I had to mentally filter out for other packages (System.Text.Encodings.Web, System.Runtime.InteropServices, and Microsoft.Bcl.asyncingterfaces). It'd be great to clean up the generation tool for that as well because if we're going to enable nuget audit and be updating packages regularly that aren't in source build, we need the process to be as smooth as possible.

mthalman (approved):

src/referencePackages/src/system.text.json/8.0.5/lib/netstandard2.0/System.Text.Json.cs
mthalman: > Also, the generate command when I ran it generated a bunch of additional files that I had to mentally filter out for other packages (System.Text.Encodings.Web, System.Runtime.InteropServices, and Microsoft.Bcl.asyncingterfaces). It'd be great to clean up the generation tool for that as well because if we're going to enable nuget audit and be updating packages regularly that aren't in source build, we need the process to be as smooth as possible.

This is called out in https://github.com/dotnet/source-build-reference-packages?tab=readme-ov-file#workflow. It is ok to include those changes.

PR #1061, opened 2024-10-16
Update SBRP self-reference

ellahathaway:


mthalman (commented):
This shouldn't be necessary since there are no newly added packages from the most recent release (eng/Build.props). Is that correct, @MichaelSimons?

ellahathaway:
Good point, I’ll just close this then.

PR #1068, opened 2024-10-29
Update runtime dependency packages to latest from maintenance-packages

carlossanlop:
Unblocks https://github.com/dotnet/runtime/pull/108806

We have generated new preview versions of some OOB packages that have been recently migrated to the maintenance-packages repo. We need to include them in sbrp so that we can update runtime's `main` dependencies to those preview versions without breaking source build.

The packages that we want to update in runtime are:
- Microsoft.Bcl.HashCode
- System.Buffers
- System.Memory
- System.Numerics.Vectors
- System.Runtime.CompilerServices.Unsafe
- System.Threading.Tasks.Extensions

I used the generate.sh script to generate these files, per the instructions.

For additional details, see the decided plan here: https://github.com/dotnet/maintenance-packages/pull/140#issuecomment-2444383987

cc @ericstj @ViktorHofer @MichaelSimons

### DIFF
--- a/NuGet.config
+++ b/NuGet.config
@@ -5,6 +5,7 @@
   </fallbackPackageFolders>
   <packageSources>
     <clear />
+    <add key="dotnet-libraries" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-libraries/nuget/v3/index.json" />
     <add key="dotnet-public" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-public/nuget/v3/index.json" />
     <add key="dotnet-tools" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-tools/nuget/v3/index.json" />
     <add key="dotnet-eng" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-eng/nuget/v3/index.json" />
--- a/eng/Build.props
+++ b/eng/Build.props
@@ -34,8 +34,8 @@
     <DependencyPackageProjects Include="$(RepoRoot)src\referencePackages\src\**\Nuget.Credentials.6.11.0.csproj" />
     <DependencyPackageProjects Include="$(RepoRoot)src\referencePackages\src\**\Nuget.ProjectModel.6.11.0.csproj" />
     <DependencyPackageProjects Include="$(RepoRoot)src\referencePackages\src\**\Nuget.Commands.6.11.0.csproj" />
-    <DependencyPackageProjects Include="$(RepoRoot)src\textOnlyPackages\src\**\Microsoft.NetCore.Platforms.3.1.0.csproj" />
     <DependencyPackageProjects Include="$(RepoRoot)src\referencePackages\src\**\System.Text.Json.8.0.5.csproj" />
+    <DependencyPackageProjects Include="$(RepoRoot)src\textOnlyPackages\src\**\Microsoft.NetCore.Platforms.3.1.0.csproj" />
     <DependencyPackageProjects Include="$(RepoRoot)src\referencePackages\src\**\System.Buffers.4.6.0-preview.1.24529.4.csproj" />
     <DependencyPackageProjects Include="$(RepoRoot)src\referencePackages\src\**\System.Numerics.Vectors.4.6.0-preview.1.24529.4.csproj" />
     <DependencyPackageProjects Include="$(RepoRoot)src\referencePackages\src\**\System.Runtime.CompilerServices.Unsafe.6.1.0-preview.1.24529.4.csproj" />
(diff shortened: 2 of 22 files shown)


MichaelSimons:
Per the [instructions for adding new packages](https://github.com/dotnet/source-build-reference-packages/blob/main/README.md#workflow): 

Add DependencyPackageProjects for all new projects in the [eng/Build.props](https://github.com/dotnet/source-build-reference-packages/blob/main/eng/Build.props#L9) in the correct dependency order.

carlossanlop:
Done. Sorry, I missed that line in the instructions.

eng/Build.props
```diff
@@ -36,6 +37,11 @@
     <DependencyPackageProjects Include="$(RepoRoot)src\referencePackages\src\**\Nuget.Commands.6.11.0.csproj" />
     <DependencyPackageProjects Include="$(RepoRoot)src\referencePackages\src\**\System.Text.Json.8.0.5.csproj" />
     <DependencyPackageProjects Include="$(RepoRoot)src\textOnlyPackages\src\**\Microsoft.NetCore.Platforms.3.1.0.csproj" />
+    <DependencyPackageProjects Include="$(RepoRoot)src\referencePackages\src\**\System.Buffers.4.6.0-preview.1.24529.4.csproj" />
+    <DependencyPackageProjects Include="$(RepoRoot)src\referencePackages\src\**\System.Memory.4.6.0-preview.1.24529.4.csproj" />
```
ViktorHofer, line 41: System.Memory depends on Buffers, Vectors and Unsafe and therefore must be listed after those.

carlossanlop:
I noticed Bcl.AsyncInterfaces is not added to Build.props, is that expected? (I did not add that one, it was already there).

carlossanlop:
Oh - We only add to DependencyPackageProjects those projects that are dependency of others. We don't add all that we add. There are many other packages in the referencePackages folder that are not in Build.props.

eng/Build.props
```diff
@@ -36,6 +36,11 @@
     <DependencyPackageProjects Include="$(RepoRoot)src\referencePackages\src\**\Nuget.Commands.6.11.0.csproj" />
     <DependencyPackageProjects Include="$(RepoRoot)src\referencePackages\src\**\System.Text.Json.8.0.5.csproj" />
     <DependencyPackageProjects Include="$(RepoRoot)src\textOnlyPackages\src\**\Microsoft.NetCore.Platforms.3.1.0.csproj" />
+    <DependencyPackageProjects Include="$(RepoRoot)src\referencePackages\src\**\System.Buffers.4.6.0-preview.1.24529.4.csproj" />
+    <DependencyPackageProjects Include="$(RepoRoot)src\referencePackages\src\**\System.Numerics.Vectors.4.6.0-preview.1.24529.4.csproj" />
+    <DependencyPackageProjects Include="$(RepoRoot)src\referencePackages\src\**\System.Runtime.CompilerServices.Unsafe.6.1.0-preview.1.24529.4.csproj" />
+    <DependencyPackageProjects Include="$(RepoRoot)src\referencePackages\src\**\System.Memory.4.6.0-preview.1.24529.4.csproj" />
+    <DependencyPackageProjects Include="$(RepoRoot)src\referencePackages\src\**\System.Threading.Tasks.Extensions.4.6.0-preview.1.24529.4.csproj" />
```
carlossanlop, line 43: Memory depends on Buffers, Vectors and Unsafe.
Threading.Tasks.Extensions depends on Unsafe.

carlossanlop:
Oh ok got it. Then I'll re-add the one I just removed, Bcl.HashCode.

ellahathaway (approved):
LGTM.

The only thing that caught my eye was the preview versions, but I see that using previous versions here was approved by @MichaelSimons (see https://github.com/dotnet/maintenance-packages/pull/140#issuecomment-2444383987) and also mentioned in the PR description.

carlossanlop:
This is all going into main, we want to test these preview versions in runtime. I am currently in the process of stabilizing the m-p repo to generate the official GA versions of these packages. If needed, I'll be happy to generate the reference packages for GA too.

NuGet.config
```diff
@@ -5,6 +5,7 @@
   </fallbackPackageFolders>
   <packageSources>
     <clear />
+    <add key="dotnet-libraries" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-libraries/nuget/v3/index.json" />
```
ViktorHofer, line 8: Should this feed get checked in?

ViktorHofer (approved):

NuGet.config
MichaelSimons: Yes it should - as long at the checked in packages come from this feed it should be checked in to support regen.  This should get cleaned up once the preview maintenance packages are removed and the feed is no longer required.

carlossanlop: The generate.sh script fails to find the packages without this feed.

PR #1099, opened 2024-12-16
Delete analyzers from 9.0 targeting packs

mthalman:
Analyzers should be excluded from the targeting packs. See https://github.com/dotnet/sdk/pull/45435#issuecomment-2546686633. Confirmed that this was already done for the 8.0 targeting packs.


README.md
```diff
@@ -90,6 +90,8 @@ You can search for known issues in the [Known Generator Issues Markdown file](do
 Generating new targeting packages is not supported. No new targeting packs should be needed/added. If you feel
 a new targeting pack is needed, please [open a new issue](#filing-issues) to discuss.
 
+When adding targeting packs, the `analyzers` directory should be excluded.
```
MichaelSimons, line 93: I think this would be better added to https://dev.azure.com/dnceng/internal/_git/dotnet-source-build-utilities?path=/IlGenerator/README.md.  I likely wouldn't see/remember this here whereas I always have to refer to the source-build-utilities readme.

MichaelSimons (approved):
One doc suggestion.  Core changes LGTM

MichaelSimons (approved):

PR #1109, opened 2025-01-09
Add MSBuild 17.12.6 packages

ViktorHofer:

PR #1113, opened 2025-01-10
Add back some recently removed packages

MichaelSimons:
Adding back the following packages:

referencePackages/src/system.formats.asn1/8.0.0
referencePackages/src/system.security.cryptography.pkcs/8.0.0
referencePackages/src/system.security.cryptography.xml/8.0.0
referencePackages/src/system.text.json/8.0.4

These packages were removed in https://github.com/dotnet/source-build-reference-packages/pull/1108 but https://github.com/dotnet/source-build-reference-packages/pull/1110 and https://github.com/dotnet/source-build-reference-packages/pull/1109 which were merged after added dependencies on them.  It was not caught in validation because the packages existed in the n-1 self-reference.  This would be solved by https://github.com/dotnet/source-build/issues/1690.

### DIFF
--- a/src/referencePackages/src/system.formats.asn1/8.0.0/System.Formats.Asn1.8.0.0.csproj
+++ b/src/referencePackages/src/system.formats.asn1/8.0.0/System.Formats.Asn1.8.0.0.csproj
@@ -1,0 +1,14 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFrameworks>net6.0;net7.0;net8.0;netstandard2.0</TargetFrameworks>
+    <AssemblyName>System.Formats.Asn1</AssemblyName>
+    <StrongNameKeyId>Open</StrongNameKeyId>
+  </PropertyGroup>
+
+  <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
+    <PackageReference Include="System.Buffers" Version="4.5.1" />
+    <PackageReference Include="System.Memory" Version="4.5.5" />
+  </ItemGroup>
+
+</Project>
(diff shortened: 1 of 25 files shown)


ellahathaway:
Orthogonal to the changes in this PR - it's unfortunate that we have to reintroduce `referencePackages/src/system.formats.asn1/8.0.0` due to its CVE, which is causing CG alerts in SBRP: https://github.com/dotnet/source-build/issues/4751.

I understand it's necessary in this case, but I wonder if we could have implemented something in PR validation to detect that the packages referencing this dependency are dependent on a package with a CVE.

NikolaMilosavljevic:
Are `DependencyPackageProjects` entries in `eng/Build.props` also needed?

MichaelSimons:
> Are DependencyPackageProjects entries in eng/Build.props also needed?

No because the packages are already in the n-1 artifacts.

NikolaMilosavljevic (approved):

ViktorHofer:
> Orthogonal to the changes in this PR - it's unfortunate that we have to reintroduce referencePackages/src/system.formats.asn1/8.0.0

System.Security.Cryptography.Pkcs/8.0.0 is the only package that references System.Formats.Asn/8.0.0. Should we patch that to reference the 8.0.1 version instead?

MichaelSimons:
> System.Security.Cryptography.Pkcs/8.0.0 is the only package that references System.Formats.Asn/8.0.0. Should we patch SSCP to reference the 8.0.1 version of SFAsn1 instead?

Yes that is the way to address the CG.  That work will be done as part of https://github.com/dotnet/source-build/issues/4751.

ViktorHofer:
I should have done that in my original PR but missed it - sorry.

PR #1134, opened 2025-01-28
Exclude reference packages from CG

MichaelSimons:
Fixes https://github.com/dotnet/source-build/issues/4838

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,19 +98,9 @@ a new targeting pack is needed, please [open a new issue](#filing-issues) to dis
 
 ## Vulnerable Packages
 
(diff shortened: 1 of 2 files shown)


mthalman (approved):
Should we also revert the manual edits to the projects to reference non-vulnerable versions?

MichaelSimons:
> Should we also revert the manual edits to the projects to reference non-vulnerable versions?

I am planning on doing that in a separate PR (I should have called it out).  I want to keep this simple to make it easy to backport.  I do not plan on backporting any regen.

PR #1138, opened 2025-01-28
Regen existing packages

MichaelSimons:
Regenerated packages that contain source-edits.  You don't see them because I re-applied the customizations after regenning the packages.  There were a few that were able to be dropped.  For example NuGet.LibraryModel.6.11.0.csproj was updated to make use of the common RequiredModifierAttributes.cs.


src/referencePackages/src/microsoft.win32.systemevents/4.7.0/microsoft.win32.systemevents.nuspec
```diff
@@ -22,8 +22,6 @@ When using NuGet 3.x this package requires at least version 3.4.</description>
     <serviceable>true</serviceable>
     <dependencies>
       <group targetFramework=".NETCoreApp3.0">
-        <!-- Manually updated to resolve 3.1.0 prebuilt.
```
MichaelSimons, line 25: I dropped this comment/customization because the code gen'd reference is not being changed.

src/referencePackages/src/system.collections.immutable/6.0.0/system.collections.immutable.nuspec
MichaelSimons, line 1: Note this is a permissions change (made by the generate tooling).  Not sure why the nuspec was defined as it was.

mthalman (approved):

PR #1139, opened 2025-01-29
Update Microsoft.Extensions.DependencyModel to 6.0.2

nohwnd:
Update the version so we can consume newer version in vstest and code coverage which depends on non-vulnerable system.text.json.


nohwnd:
In general what is the policy here about updating the packages? there are much newer versions 8.0.0 and 9.0.0 that we would also be able to use, and might prefer to use. When we updated vstest, few packages in here were older than expected. Can I just identify them and post PRs, or do you prefer to keep this repo with as few versions of packages as possible?

nohwnd:
fyi @fhnaseer

MichaelSimons:
> In general what is the policy here about updating the packages? there are much newer versions 8.0.0 and 9.0.0 that we would also be able to use, and might prefer to use. When we updated vstest, few packages in here were older than expected. Can I just identify them and post PRs, or do you prefer to keep this repo with as few versions of packages as possible?

You are free to open PRs and add as many versions that are needed.  The goal of this repo is to support what ever versions the repos want to use via a self service model.  The source-build team as maintainers will periodically remove unused packages so there is no need for repo owners to worry about cleanup.

src/referencePackages/src/system.runtime.interopservices/4.3.0/System.Runtime.InteropServices.4.3.0.csproj
```diff
@@ -43,9 +40,6 @@
   </ItemGroup>
 
   <ItemGroup Condition="'$(TargetFramework)' == 'netstandard1.5'">
-    <!-- Needs additional APIs that don't exist in the contract. -->
```
MichaelSimons, line 46: You are reverting some intentional tweaks to the generated code.  The changes to this file as well as the deletion of https://github.com/dotnet/source-build-reference-packages/pull/1139/files#diff-217e4e4d9fa100fb4df1c2ff5c488dc55e6e4c9c6fe15f4c9b4c7bce389bac72 will need to be reverted.

nohwnd: reverted.

src/referencePackages/src/system.runtime.interopservices/4.3.0/System.Runtime.InteropServices.Manual.cs
MichaelSimons, line 1: This file deletion needs to be reverted.

src/referencePackages/src/system.buffers/4.5.1/ref/netstandard1.1/System.Buffers.cs
```diff
@@ -4,7 +4,6 @@
 // ------------------------------------------------------------------------------
 // Changes to this file must follow the http://aka.ms/api-review process.
 // ------------------------------------------------------------------------------
-[assembly: System.Reflection.AssemblyMetadata("source", "source-build-reference-packages")]
```
MichaelSimons, line 7: I just realized that you must be running with an older toolset locally.  You should clean your repo to delete you local artifacts directory and .dotnet.  These attributes were added some time ago now.

MichaelSimons, line 7: I just realized that you must be running with an older toolset locally.  You should clean your repo to delete you local artifacts directory and .dotnet.  These attributes were added some time ago now.

MichaelSimons: I just discovered there is a difference in the generated code between running the generator tools and windows vs linux.  This is not good.  I will log and issue and fixup this PR.

eng/Build.props
```diff
@@ -39,6 +39,10 @@
     <DependencyPackageProjects Include="$(RepoRoot)src\referencePackages\src\**\System.Text.Json.6.0.0.csproj" />
     <DependencyPackageProjects Include="$(RepoRoot)src\referencePackages\src\**\System.Text.Json.8.0.4.csproj" />
 
+    <DependencyPackageProjects Include="$(RepoRoot)src\referencePackages\src\**\System.Text.Encodings.Web.6.0.1.csproj" />
```
MichaelSimons, line 42: Friendly reminder to add these DependencyPackageProjects whenever adding new packages.  This is documented [here](https://github.com/dotnet/source-build-reference-packages/blob/main/README.md#workflow).

src/referencePackages/src/system.runtime.interopservices/4.3.0/System.Runtime.InteropServices.Manual.cs
MichaelSimons: I fixed this.

src/referencePackages/src/system.buffers/4.5.1/ref/netstandard1.1/System.Buffers.cs
MichaelSimons: It appears that there is a difference between the generated code on Windows versus Linux.  I will open an issue on this.  I regen'd everything on linux which reverted all of the updated files.

nohwnd: thank you. please link the issue, or ping me one you have it please. I don't see issues on this repo. I will use wsl until that is fixed when I need to make more changes. We plan to update all packages on vstest to latest, so I will have to add few here.

MichaelSimons (approved):

src/referencePackages/src/system.buffers/4.5.1/ref/netstandard1.1/System.Buffers.cs
MichaelSimons: https://github.com/dotnet/source-build/issues/4858

PR #1140, opened 2025-01-29
Add infra to preserve customizations on package regen

MichaelSimons:
Introduced two new constructs to encapsulated customizations to the generated artifacts.

  * Customizations.props - Automatically imported by the generated project. Use it for additive changes such as NoWarns or additional source files.
  * Customizations.cs - This is not automatically included by the generated project. Use it to add new types or members to partial classes.

The benefit of using these files is that they will be preserved when the packages are regenerated.  These constructs are not intended to support all scenarios e.g. add an attribute to a property.

I am open to suggestions on the names of these constructs.

A number of projects were updated to utilize these new constructs.


ViktorHofer:
We briefly talked about this in chat. Manual edits to NuGet.config are hard to isolate into separate files but there don't seem to be many. For the Customizations.cs, this definitely helps with new types or new members. Changing member is unfortunately not possible unless we tell GenAPI to exclude certain members (which is possible) and then add them to that secondary file. I don't think that's a good use here though as it would mean that we would need to list those symbols separately for GenAPI.

Aside from that, this looks good and will definitely help with re-generating packages.

src/referencePackages/Directory.Build.props
```diff
@@ -1,7 +1,13 @@
 <Project>
 
   <Import Project="..\..\Directory.Build.props" />
-  
+
+  <PropertyGroup>
+    <CustomizationsPropsPath>$(MSBuildProjectDirectory)\$(CustomizationsPropsFile)</CustomizationsPropsPath>
+  </PropertyGroup>
```
ViktorHofer, line 7: What's the reason for not also automatically including Customizations.cs if that file exists?

src/referencePackages/src/system.runtime.interopservices/4.1.0/Customizations.props
```diff
@@ -0,0 +1,6 @@
+<Project>
+  <ItemGroup Condition="'$(TargetFramework)' == 'netcoreapp1.1' or '$(TargetFramework)' == 'netstandard1.5'">
+    <!-- Needs additional APIs that don't exist in the contract. -->
+    <Compile Include="$(MSBuildThisFileDirectory)Customizations.cs" />
+  </ItemGroup>
```
ViktorHofer, line 5: It might be easier to just always import Customizations.cs files if they exist and put these TFM checks directly into that file. The roslyn repo often prefers that over conditioning in the project file and that worked great for them.

This would mean that the Customizations.cs would start with a `#if NETCOREAPP1_1 || NETSTANDARD1_5` line.

MichaelSimons: Hmm, I originally thought this pattern would provide more flexibility but I don't think that is necessarily the case.  The pattern you suggest reduces the onboarding maintenance/ease of use.  I will make this change.

src/referencePackages/Directory.Build.props
MichaelSimons: Discussed in your other comment.

src/referencePackages/src/system.runtime.interopservices/4.1.0/Customizations.props
MichaelSimons: Addressed in https://github.com/dotnet/source-build-reference-packages/pull/1140/commits/394218c4ceb60e337d4076883846f791088f9444

ViktorHofer (approved):

mthalman (approved):
This is pretty slick.

PR #1149, opened 2025-02-04
Enable Nuget Audit

MichaelSimons:
Fixes https://github.com/dotnet/source-build/issues/4673.

Enabled the NuGet Audit functionality for the SBRP repo but disabled it for the reference and targeting packages given they contain no implementation.  It is enabled for the generate tooling as well as the text only packages.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -23,6 +23,9 @@
     <CustomizationsPrefix>Customizations</CustomizationsPrefix>
     <CustomizationsPropsFile>$(CustomizationsPrefix).props</CustomizationsPropsFile>
     <CustomizationsSourceFile>$(CustomizationsPrefix).cs</CustomizationsSourceFile>
+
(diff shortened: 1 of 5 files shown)


ViktorHofer (approved):
Consider setting this so that local builds aren't blocked when there's a new advisory: https://github.com/dotnet/arcade/blob/5d30877b1eb318583acd5f2cd3f3a80e77b3add2/Directory.Build.props#L17-L18

mthalman (approved):

MichaelSimons:
> LGTM. Consider also setting this so that local builds aren't blocked when there's a new advisory: https://github.com/dotnet/arcade/blob/5d30877b1eb318583acd5f2cd3f3a80e77b3add2/Directory.Build.props#L17-L18

WarningsAsErrors was not enabled anywhere.  It doesn't make sense to enable WarningsAsErrors for any of the ref packs but it does make sense for the generate tooling so I enabled it there.

PR #1175, opened 2025-02-25
Update SBRP self-reference with Preview 1

NikolaMilosavljevic:
Update with reference from [Preview 1 release](https://github.com/dotnet/dotnet/releases/tag/v10.0.0-preview.1.25080.5).


MichaelSimons:
I don't think we want to accept this.  If we are re-bootstrapping, we can't to this.  The self-reference must match the n-1 artifacts.

PR #1179, opened 2025-02-28
Remove netstandard1.x packages and support infrastructure

ViktorHofer:
Fixes https://github.com/dotnet/source-build/issues/4482

### Commit 1
Remove netstandard1.x only packages that weren't removed in f0ea129f0c02b5af4f096169500e10473d100231. My guess is that they weren't reported because `Microsoft.NetCore.Targets/1.1.0` depended on them (which also gets removed now as part of this commit). So there was a cycle which hindered the tooling to flag all those packages.

1. I manually removed all the package folders that were netstandard1.x only.
2. I then updated all projects that targeted netstandard1.x by removing the TFM string from the TargetFrameworks property and removed item groups conditioned on the TFM.
3. Updated the nuspec to remove the group TFM entries
4. Deleted the sources for the removed TFMs

I forgot to mention that this PR also removes a few netcoreapp1.x leftover packages which have the same matching `4.3.0` version as the netstandard1.x packages.

### Commit 2

Filter out netstandard1.x TFMs when generating new packages with the PackageSourceGenerator tooling.


ViktorHofer:
Apparently needs https://github.com/dotnet/source-build-reference-packages/pull/1180

ViktorHofer:
OK I only later noticed that there's a pipeline to create such a PR automatically. On the positive side, the pipeline wouldn't have delete the non-false-positive text-only packages which are important to delete now to get rid of netstandard1.x packages.

ViktorHofer:
**NO merge until we validated this with a full VMR source-only build**

ViktorHofer:
OK after talking to @MichaelSimons offline, he let me know that this PR removes a few packages that are still used and aren't reported correctly by the tooling: https://github.com/dotnet/source-build/issues/4924

He also suggested to wait until https://github.com/dotnet/source-build-reference-packages/pull/1181 is merged to make this PR better focused on the netstandard1.x removal.

MichaelSimons:
I queued a full VMR test build at https://github.com/dotnet/sdk/pull/47172

PR #1189, opened 2025-03-05
Normalize casing of targeting pack projects and nuspecs

ViktorHofer:
All SBRP projects except the targeting packs consistently use the casing of the package id (+ "." + packageversion), i.e.
`Microsoft.Extensions.Logging.8.0.0.csproj`.

A few targeting pack projects didn't follow that casing. Similar for the nuspecs which should always be lower-cased.

This is in preparation to use the live targeting packs in SBRP and remove the self SBRP dependency. **This change is repo-local and doesn't impact the published SBRP artifacts.**


mthalman (commented):
Thoughts on updating the directory name to match this casing?

ViktorHofer:
> Thoughts on updating the directory name to match this casing?

The directory name is expected to be lower-case. I.e. see https://github.com/dotnet/source-build-reference-packages/tree/FixCasingOfTargetingPacks/src/referencePackages/src

PR #1191, opened 2025-03-05
Various infra improvements

ViktorHofer:
- Simplify FilterImplicitAssemblyReferences target
- Add comments to a few settings
- Add missing aspnetcore KnownFrameworkReference targeting pack setting
- Disable P2Ps getting copied to a lib's output which makes the build significantly faster.
- Move the _FindDependencies property from props to targets and extend the comment on why the setting is necessary.

### DIFF
--- a/src/referencePackages/Directory.Build.targets
+++ b/src/referencePackages/Directory.Build.targets
@@ -3,9 +3,21 @@
   <Import Project="..\..\Directory.Build.targets" />
 
   <PropertyGroup>
+    <!-- Disable RAR from transitively discovering dependencies for references. This is required as we don't copy
+         dependencies over into the output directory which means RAR can't resolve them.
+         This is also necessary to make P2Ps with same package ids but different versions work. -->
+    <_FindDependencies>false</_FindDependencies>
+
     <CustomizationsPropsPath>$(MSBuildProjectDirectory)\$(CustomizationsPropsFile)</CustomizationsPropsPath>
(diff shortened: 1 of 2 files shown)


src/referencePackages/Directory.Build.targets
```diff
@@ -21,23 +33,24 @@
...
   <ItemGroup>
     <KnownFrameworkReference Update="@(KnownFrameworkReference->WithMetadataValue('Identity', 'Microsoft.NETCore.App'))">
       <TargetingPackVersion Condition="'%(TargetFramework)' == 'netcoreapp3.0'">3.0.0</TargetingPackVersion>
       <TargetingPackVersion Condition="'%(TargetFramework)' == 'netcoreapp3.1'">3.1.0</TargetingPackVersion>
       <TargetingPackVersion Condition="$([System.Text.RegularExpressions.Regex]::IsMatch('%(TargetFramework)', '^net\d+\.0$'))">$([System.Text.RegularExpressions.Regex]::Match('%(TargetFramework)', '\d+').Value).0.0</TargetingPackVersion>
     </KnownFrameworkReference>
 
+    <KnownFrameworkReference Update="@(KnownFrameworkReference->WithMetadataValue('Identity', 'Microsoft.AspNetCore.App'))">
+      <TargetingPackVersion Condition="'%(TargetFramework)' == 'net6.0'">6.0.2</TargetingPackVersion>
+    </KnownFrameworkReference>
```
mthalman, line 52: We also have a 6.0.2 version of Microsoft.NetCore.App.Ref. Does that need an entry like this as well?

MichaelSimons (approved):

PR #1193, opened 2025-03-06
Remove unreferenced Microsoft.NetCore.App.Ref 3.1.0 targeting pack

MichaelSimons:
This is no longer being referenced so it should be removed.


MichaelSimons:
Validated with https://github.com/dotnet/sdk/pull/47293.

NikolaMilosavljevic (approved):

mthalman (approved):

PR #1199, opened 2025-03-14
Remove analyzers from aspnetcore targeting pack

ViktorHofer:
SBRP targeting packs can't have analyzers as those execute and SBRP packages are contract only.

Unblocks https://github.com/dotnet/sdk/pull/47509


ViktorHofer:
@dotnet/source-build these targeting packs get auto-generated via an internal tooling. Can that get updated to not include the "analyzers" folder and also trim out the"/analyzer*" entries from the FrameworkList.xml?

MichaelSimons:
> @dotnet/source-build these targeting packs get auto-generated via an internal tooling. Can we update that tooling to not include the "analyzers" folder (if that's not already the case) and also trim out the"/analyzer*" entries from the FrameworkList.xml?

@ViktorHofer, would you mind logging an issue for this.  It is called out in the documentation for adding targeting packs but should be handled by the tooling.

ViktorHofer:
https://github.com/dotnet/source-build/issues/4949

PR #1203, opened 2025-03-19
[release/9.0] Add maintenance-packages entries

carlossanlop:
Manual backport of https://github.com/dotnet/source-build-reference-packages/pull/1200 and https://github.com/dotnet/source-build-reference-packages/pull/1151

The packages are already in nuget.org.

Backporting because several repos are still using release/9.0 as their SBRP source-branch in main: 
- fsharp https://github.com/dotnet/fsharp/pull/18329
- msbuild https://github.com/dotnet/msbuild/pull/11457
- razor https://github.com/dotnet/razor/pull/11507

The System.Runtime.InteropServices modification was made by generate.sh.

@MichaelSimons @ericstj @ViktorHofer

PR #1205, opened 2025-03-20
[main] Add maintenance-packages entries

carlossanlop:
Same as https://github.com/dotnet/source-build-reference-packages/pull/1203 but for main and without Build.props.

These packages include the recent fixes from today.

Note that System.Runtime.CompilerServices.Unsafe is not in this PR compared to the 9.0, because that one did not get a new version (and we had merged it previously already).

cc @ViktorHofer @ericstj @MichaelSimons

### DIFF
--- a/src/referencePackages/src/system.buffers/4.6.1/System.Buffers.4.6.1.csproj
+++ b/src/referencePackages/src/system.buffers/4.6.1/System.Buffers.4.6.1.csproj
@@ -1,0 +1,9 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFrameworks>netstandard2.0</TargetFrameworks>
+    <AssemblyName>System.Buffers</AssemblyName>
+    <StrongNameKeyId>Open</StrongNameKeyId>
+  </PropertyGroup>
+
+</Project>
--- a/src/referencePackages/src/system.buffers/4.6.1/System.Buffers.4.6.1.csproj
+++ b/src/referencePackages/src/system.buffers/4.6.1/System.Buffers.4.6.1.csproj
@@ -2,6 +2,7 @@
 
   <PropertyGroup>
     <TargetFrameworks>netstandard2.0</TargetFrameworks>
+    <PackageVersion>4.6.1</PackageVersion>
     <AssemblyName>System.Buffers</AssemblyName>
     <StrongNameKeyId>Open</StrongNameKeyId>
   </PropertyGroup>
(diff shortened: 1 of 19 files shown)


ViktorHofer:
I think SBRP needs a fx first. It's not respecting placeholder files and dropping TFMs for it.

ViktorHofer:
> /__w/1/s/.packages/microsoft.dotnet.arcade.sdk/10.0.0-beta.25164.6/tools/SourceBuild/AfterSourceBuild.proj(83,5): error : 7 new pre-builts discovered! Detailed usage report can be found at /__w/1/s/artifacts/sb/prebuilt-report/baseline-comparison.xml.

@MichaelSimons this error is fun. The 5 listed packages that are getting added via this PR are reported as prebuilts :D

> System.Reflection.Emit.4.7.0
System.Reflection.Emit.ILGeneration.4.7.0

@carlossanlop these two packages need to get added to SBRP as well as they are dependencies of DispatchProxy.

MichaelSimons:
> @MichaelSimons this error is fun. The 7 listed packages that are getting added via this PR are reported as prebuilts :D

That's because you can't port the source between branches now that main vs servicing define the project references differently (project vs package reference).  It is fortunate that the build order happened to catch this because it is not guaranteed.  I am going to call this out in the readme.

ViktorHofer (approved):

carlossanlop:
Thanks for regenerating the packages. Good to know I shouldn't backport.

MichaelSimons (approved):

MichaelSimons:
@carlossanlop - Can you tell me if all of the dependency work on this PR been submitted?  I ask because I am in the process of removing all SBRP that are detected as being unreferenced in the product source build.  Two of the packages added here are detected as unreferenced.  I am wondering if that is true or if they are undetected because the dependency work hasn't flown into the VMR yet.

```
/src/referencePackages/src/system.net.websockets.websocketprotocol/5.1.2
src/referencePackages/src/system.reflection.dispatchproxy/4.8.2
```

carlossanlop:
> @carlossanlop - Can you tell me if all of the dependency work on this PR been submitted? I ask because I am in the process of removing all SBRP that are detected as being unreferenced in the product source build. Two of the packages added here are detected as unreferenced. I am wondering if that is true or if they are undetected because the dependency work hasn't flown into the VMR yet.

Hey, I was out the last couple of days.

I don't think we haven't flowed everything to the VMR, there were various deps flow conflicts and we're still going through them. Let's come back to this in a couple of days.

PR #1208, opened 2025-03-24
Remove maintenance-packages temp feed from NuGet.config

carlossanlop:
The packages that were published in that feed are already available in nuget.org.

@MichaelSimons @ViktorHofer @ericstj


MichaelSimons (approved):
Thanks for cleaning this up!

PR #1212, opened 2025-03-31
Add 9.0 packages used by roslyn

ericstj:
Adding packages used after https://github.com/dotnet/roslyn/commit/22e1b3f73be8941476c7e9d8e25de1787d9ac398

See https://github.com/dotnet/runtime/pull/112937#discussion_r2021884266


MichaelSimons:
> Hmm, maybe not after reading @MichaelSimons comment - maybe we instead make these allowed pre-builts.

Yes, that is my [recommendation](https://github.com/dotnet/runtime/pull/112937#discussion_r2022951143).  If there are added here, they will get detected as unreferenced in the product source-build and will be cleaned up.

PR #1218, opened 2025-04-04
[9.0] Add prebuilts for maintenance-packages

carlossanlop:
Similar to https://github.com/dotnet/source-build-reference-packages/pull/1217 but not a backport.

- System.Memory 4.6.3
- System.Net.WebSockets.WebSocketProtocol 5.1.3
- System.Runtime.CompilerServices.Unsafe 6.1.2
- System.Threading.Tasks.Extensions 4.6.3

@MichaelSimons please let me know if the order in Build.props is correct for when we have multiple versions of the same package.


MichaelSimons:
> @MichaelSimons please let me know if the order in Build.props is correct for when we have multiple versions of the same package.

The primary requirement is that they are listed in dependency order.  Secondary, they should be grouped by when they are added.  This aides in cleanup after a rebootstrap.  I pushed small change to group them.

MichaelSimons (approved):

PR #1221, opened 2025-04-09
[main] Add Nuget 6.12.1 packages

v-wuzhai:


eng/Build.props
```diff
@@ -13,6 +13,14 @@
     <ProjectToBuild Include="$(RepoRoot)src\targetPacks\ILsrc\**\*.csproj" BuildInParallel="false" />
     <ProjectToBuild Include="$(RepoRoot)src\textOnlyPackages\src\*\*\*.csproj" />
     <ProjectToBuild Include="$(RepoRoot)src\referencePackages\src\**\*.csproj" />
+
```
MichaelSimons, line 16: These are no longer necessary in main and should be removed.

MichaelSimons:
There are a few customizations that are necessary to the generated artifacts for nuget.  My suggestion is to compare the generated artifacts to the existing 6.13 version to see what they are.

MichaelSimons (approved):

src/referencePackages/src/system.text.json/8.0.4/System.Text.Json.8.0.4.csproj
ViktorHofer, line 1: I'm not sure what the current policy for SBRP is but the upstream STJ/8.0.4 is marked as vulnerable and shouldn't be used. cc @MichaelSimons

mthalman: Component detection is disabled for SBRP since it contains no implementation. It's up to the consuming product repos to address any dependency on vulnerable versions.

PR #1227, opened 2025-04-18
Add 4.0.1 packages for Microsoft.CodeAnalysis

mthalman:
This is necessary to support dev builds of source build: https://github.com/dotnet/source-build/issues/5061. Without this, it leads to package downgrade errors like the following:

```
/repos/dotnet/src/roslyn/src/RoslynAnalyzers/Microsoft.CodeAnalysis.ResxSourceGenerator/Microsoft.CodeAnalysis.ResxSourceGenerator.VisualBasic/Microsoft.CodeAnalysis.ResxSourceGenerator.VisualBasic.csproj error NU1109: Detected package downgrade: System.Collections.Immutable from 10.0.0-preview.4.25211.17 to centrally defined 10.0.0-dev. Update the centrally managed package version to a higher version. 
 *Microsoft.CodeAnalysis.ResxSourceGenerator.VisualBasic.csproj* -> Microsoft.CodeAnalysis.Common 5.0.0-1.25211.1 -> System.Collections.Immutable (>= 10.0.0-preview.4.25211.17) 
 *Microsoft.CodeAnalysis.ResxSourceGenerator.VisualBasic.csproj* -> System.Collections.Immutable (>= 10.0.0-dev) [/repos/dotnet/src/roslyn/Roslyn.sln]
```
Binlog: [source-inner-build.roslyn.downgrade.zip](https://github.com/user-attachments/files/19813486/source-inner-build.roslyn.downgrade.zip)

The intent here is to change these roslyn analyzers to depend on the [exact Microsoft.CodeAnalysis version defined in Version.props](https://github.com/dotnet/roslyn/blob/72266a2834d24086e9499e0ad49eaa8680dd0c6c/eng/Versions.props#L138) rather than lifting it. That will prevent the downgrade from happening.

When reviewing, look through the commits. There were some manual changes that were necessary to get this to work:
* Removed Workspaces dependency on Humanizer (this will be provided by SBE)
* Workspaces had a dependency on very old version of System.Composition which only provided an unsupported netstandard TFM. Upgraded to 9.0 version which aligns with the version used by Roslyn.
* Generated code fixes


src/referencePackages/src/microsoft.codeanalysis.csharp.workspaces/4.0.1/Customizations.props
```diff
@@ -0,0 +1,8 @@
+<Project>
+
+  <ItemGroup>
+    <!-- Humanizer is provided through source-build-externals -->
+    <ProjectReference Remove="../../humanizer.core/2.2.0/Humanizer.Core.2.2.0.csproj" />
```
MichaelSimons, line 5: Would it make sense to replace this with a package dependency instead?  The primary benefit is to ensure the dependency exists to know that the resulting nuget package can be restored without prebuilts.  This applied to all places using this pattern.

src/referencePackages/src/microsoft.codeanalysis.workspaces.common/4.0.1/Customizations.props
```diff
@@ -0,0 +1,16 @@
+<Project>
+
+  <ItemGroup>
+    <!-- Humanizer is provided through source-build-externals -->
+    <ProjectReference Remove="../../humanizer.core/2.2.0/Humanizer.Core.2.2.0.csproj" />
+
+    <!-- Upgrade System.Composition to a version that has netstandard2.0 -->
```
MichaelSimons, line 7: This comment should clarify why 9.0.0 versus the minimum version that support netstandard2.0.  I think it would be helpful to call out that the metadata is eliminated.

MichaelSimons (commented):
It would be nice to avoid the System.Composition lifting but I think what you have done is reasonable.

src/referencePackages/src/system.io.pipelines/5.0.1/System.IO.Pipelines.5.0.1.csproj
```diff
@@ -0,0 +1,16 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFrameworks>netcoreapp3.0;netstandard2.0</TargetFrameworks>
```
MichaelSimons, line 4: This should be regenerated after https://github.com/dotnet/source-build-reference-packages/pull/1228 is merged.

MichaelSimons: Regenerated in https://github.com/dotnet/source-build-reference-packages/pull/1227/commits/9d756e26e7e786b4e209ac8796dd51133d66da48

eng/Versions.props
```diff
@@ -8,6 +8,7 @@
     <UsingToolXliff>false</UsingToolXliff>
   </PropertyGroup>
   <PropertyGroup>
+    <HumanizerCoreVersion>2.14.1</HumanizerCoreVersion>
```
MichaelSimons, line 11: Do we want to automatically lift this version if SBE is updated or should we stay on the version declared by M.CA?  There is precedence for staying on the declared versions.  Sys.Comp is the only exception.

mthalman: Wouldn't it cause a prebuilt to reference the originally declared version?

src/referencePackages/src/microsoft.codeanalysis.workspaces.common/4.0.1/Customizations.props
```diff
@@ -0,0 +1,20 @@
+<Project>
+
+  <ItemGroup>
+    <!-- Humanizer is provided through source-build-externals -->
+    <ProjectReference Remove="../../humanizer.core/2.2.0/Humanizer.Core.2.2.0.csproj" />
+
```
MichaelSimons, line 6: This should have a package reference added to the one provided by SBE.

MichaelSimons (approved):

PR #1074, opened 2024-11-08
Remove SBRP solution file

Winniexu01:
Fixes https://github.com/dotnet/source-build/issues/4716

This solution file is added by https://github.com/dotnet/source-build-reference-packages/pull/1036. It seems unnecessary, so I removed it.

PR #1230, opened 2025-04-23
Remove SBRP solution file

Winniexu01:
Fixes https://github.com/dotnet/source-build/issues/5069

This solution file is added by https://github.com/dotnet/source-build-reference-packages/pull/1227.

Related to https://github.com/dotnet/source-build-reference-packages/pull/1074.


akoeplinger:
do we know what causes this file to get added all the time?

PR #1262, opened 2025-06-03
[DRAFT] VMR Verification in SBRP PRs

NikolaMilosavljevic:
Verification of changes needed for source-only runs of the VMR build pipeline.

### DIFF
--- a/azure-pipelines/builds/vmr-build-pr.yml
+++ b/azure-pipelines/builds/vmr-build-pr.yml
@@ -16,6 +16,7 @@ resources:
     type: github
     name: dotnet/dotnet
     endpoint: dotnet
+    ref: refs/heads/vmrverification.list
 
 stages:
 - template: /eng/pipelines/templates/stages/vmr-build.yml@vmr


NikolaMilosavljevic:
Hmm, default value for `isSourceOnlyBuild` is `false`. Not sure if it's possible to specify a different value of the `source-build-reference-package-unified-build` pipeline.

The alternative would be to rename this pipeline to be `source-build-reference-packages-source-build`, and update the template to set `isSourceOnlyBuild` based on the pipeline name, similar to how we had it in old `sdk` pipelines/templates: https://github.com/dotnet/sdk/blob/9dee3cf508254bc3859d9cd2fe96aeb2ad1b9751/eng/pipelines/vmr-build-pr.yml#L56-L58

@MichaelSimons

NikolaMilosavljevic:
> > Hmm, default value for `isSourceOnlyBuild` is `false`. Not sure if it's possible to specify a different value of the `source-build-reference-package-unified-build` pipeline.
> > The alternative would be to rename this pipeline to be `source-build-reference-packages-source-build`, and update the template to set `isSourceOnlyBuild` based on the pipeline name, similar to how we had it in old `sdk` pipelines/templates: https://github.com/dotnet/sdk/blob/9dee3cf508254bc3859d9cd2fe96aeb2ad1b9751/eng/pipelines/vmr-build-pr.yml#L56-L58
> > Another option would be to have a different template for source-only builds. Pipelines would reference the template that is most appropriate for the repo, i.e. all legs, source-only, Windows-only, etc.
> > @MichaelSimons
> 
> I'm not fond of a name convention based approach. It is not intuitive and doesn't provide a very good UX for changing the configuration down the road. It would be nice to have a scalable solution that would solve all future scenarios e.g. Windows only legs, single linux leg, etc. plus a combination of configurations
> 
> Could the parameters be changed to variables which can then be saved in the AzDO pipeline configuration?

Thanks - yes we could use variables instead - will test it out, and expand for other scenarios.

mmitche:
Another option (already discussed) would be to, rather than having eng/common/vmr-build-pr.yml be the root YAML for the job, have a repo implement a stub. The stub would have the triggers and call vmr-build-pr.yml, specifying that this is a repo that only needs source-build legs.

> Runtime variable can be used to conditionally include a stage or template - option 2 above. But, cannot be converted to boolean in a compile-time conversion, to be passed in to a template. This is a really unfortunate constraint.

Is it possible to use the conditional include of a template to include a variables template which defines boolean typed variables, which then get used? Therefore you're not actually converting the variable directly to a boolean.

NikolaMilosavljevic:
> > @MichaelSimons @ViktorHofer any opinion on this approach? It makes pipeline yml and intent more clear and descriptive, and any modification becomes straightforward.
> 
> I think this is worthwhile. I would like to see what this would look like.

Here's how the changes would look: https://github.com/dotnet/source-build-reference-packages/commit/c51f6bf70a890c5da2575acc451d80beccf7fa11

With this approach, we would be adding a requirement for another repo-specific YML. Various elements need to be moved to this new YML, including `trigger:`, `pr:`, `variables:` and `resources:`

This would diminish the value of the current shared pipeline YML. The benefit is that we can use YML `boolean` parameters to condition VMR verification. The sample isn't complete, but I have added a, currently unused, parameter for conditioning VMR for Windows builds.

@MichaelSimons @mmitche

NikolaMilosavljevic:
> > > > @MichaelSimons @ViktorHofer any opinion on this approach? It makes pipeline yml and intent more clear and descriptive, and any modification becomes straightforward.
> > > 
> > > 
> > > I think this is worthwhile. I would like to see what this would look like.
> > 
> > 
> > Here's how the changes would look: [c51f6bf](https://github.com/dotnet/source-build-reference-packages/commit/c51f6bf70a890c5da2575acc451d80beccf7fa11)
> > With this approach, we would be adding a requirement for another repo-specific YML. Various elements need to be moved to this new YML, including `trigger:`, `pr:`, `variables:` and `resources:`
> > This would diminish the value of the current shared pipeline YML. The benefit is that we can use YML `boolean` parameters to condition VMR verification. The sample isn't complete, but I have added a, currently unused, parameter for conditioning VMR for Windows builds.
> > @MichaelSimons @mmitche
> 
> The root pipeline is still a template that is easy to copy between repos, so no concerns there.
> 
> One suggestion is that I would avoid going with booleans to conditionalize the VMR PR testing that is performed. Instead, long term it may be cleaner to pass this info as a list. e.g.:
> 
> ```
> parameters:
>   - name: verifications
>     type: object
>     default: []
> 
> jobs:
> 
> # Job for A
> - ${{ if containsValue(parameters.verifications, 'source-build') }}:
>   - job: TestA
>     steps:
>     - script: echo "Do source-build check"
> ```

Yeah, I agree with a list of verifications, and the intent to use these types of checks for PR verification jobs only. The alternative would be to use these conditions for every job we run, which would make YMLs much harder to read.

However, `verifications` parameter needs to have a default value that enables running all of the current jobs, to not regress the VMR build.

This would imply that we need the following default list of items:: `unified-build`, `source-build`, `windows`, `linux`, `ios`, `browser`, `android`. The alternative would be: `unified-build-linux`, `unified-build-windows`, etc.

`verifications` parameter, with correct set of default values, needs to exist in both VMR template (`vmr-build.yml` stage) and in `arcade`'s vmr-build-pr.yml`. This adds maintenance cost, to keep these in sync.

We'd end up with something list this:

```
jobs:

# Job for android - unified-build
- ${{ if and(containsValue(parameters.verifications, 'unified-build'), containsValue(parameters.verifications, 'android')) }}:
  - template: ../jobs/vmr-build.yml
    ...

# Job for Windows - unified-build
- ${{ if and(containsValue(parameters.verifications, 'unified-build'), containsValue(parameters.verifications, 'windows')) }}:
  - template: ../jobs/vmr-build.yml
    ...
```

mmitche:
I don't think you'd need the default list to exist in both templates. You'd set the default in the outer template to something like `default`, then pass that to the inner template. If the inner template receives `default` then the default list is used.

NikolaMilosavljevic:
> I don't think you'd need the default list to exist in both templates. You'd set the default in the outer template to something like `default`, then pass that to the inner template. If the inner template receives `default` then the default list is used.

Hmm, never used that before. Is `default` a magic value telling Azdo preprocessor to use the default values in the inner template? Or would I need to check for this value and initialize some inner value (perhaps variable) based on the supplied value of this parameter?

What we will have are 3 levels of YMLs:

`repo-specific.yml`
```
- template arcade-shared.yml
  parameters:
    verifications: source-build
```

`arcade-shared.yml`
```
parameters:
- name: verifications
  type: object
  default:
    source-build
    unified-build
    ...etc...

- template vmr-template.yml@vmr
  parameters:
    verifications: {{ parameters.verifications }}
```

`vmr-template.yml`
```
parameters:
- name: verifications
  type: object
  default:
    source-build
    unified-build
    ...etc...

jobs:
# Job for android - unified-build
- ${{ if and(containsValue(parameters.verifications, 'unified-build'), containsValue(parameters.verifications, 'android')) }}:
  - template: ../jobs/vmr-build.yml
```

`vmr-template.yml` definitely needs to have a default list of values, to allow VMR to build all default verification jobs.

mmitche:
> > I don't think you'd need the default list to exist in both templates. You'd set the default in the outer template to something like `default`, then pass that to the inner template. If the inner template receives `default` then the default list is used.
> 
> Hmm, never used that before. Is `default` a magic value telling Azdo preprocessor to use the default values in the inner template? Or would I need to check for this value and initialize some inner value (perhaps variable) based on the supplied value of this parameter?
> 
> What we will have are 3 levels of YMLs:
> 
> `repo-specific.yml`
> 
> ```
> - template arcade-shared.yml
>   parameters:
>     verifications: source-build
> ```
> 
> `arcade-shared.yml`
> 
> ```
> parameters:
> - name: verifications
>   type: object
>   default:
>     source-build
>     unified-build
>     ...etc...
> 
> - template vmr-template.yml@vmr
>   parameters:
>     verifications: {{ parameters.verifications }}
> ```
> 
> `vmr-template.yml`
> 
> ```
> parameters:
> - name: verifications
>   type: object
>   default:
>     source-build
>     unified-build
>     ...etc...
> 
> jobs:
> # Job for android - unified-build
> - ${{ if and(containsValue(parameters.verifications, 'unified-build'), containsValue(parameters.verifications, 'android')) }}:
>   - template: ../jobs/vmr-build.yml
> ```
> 
> `vmr-template.yml` definitely needs to have a default list of values, to allow VMR to build all default verification jobs.

You'd check for the value. I've used this pattern a few times in the VMR. The outer pipeline has a parameter for signing, say, with a string "Default to whatever the default for this branch is". Then the variables template checks for that value and re-interprets it as the appropriate value.

NikolaMilosavljevic:
> > > I don't think you'd need the default list to exist in both templates. You'd set the default in the outer template to something like `default`, then pass that to the inner template. If the inner template receives `default` then the default list is used.
> > 
> > 
> > Hmm, never used that before. Is `default` a magic value telling Azdo preprocessor to use the default values in the inner template? Or would I need to check for this value and initialize some inner value (perhaps variable) based on the supplied value of this parameter?
> > What we will have are 3 levels of YMLs:
> > `repo-specific.yml`
> > ```
> > - template arcade-shared.yml
> >   parameters:
> >     verifications: source-build
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > `arcade-shared.yml`
> > ```
> > parameters:
> > - name: verifications
> >   type: object
> >   default:
> >     source-build
> >     unified-build
> >     ...etc...
> > 
> > - template vmr-template.yml@vmr
> >   parameters:
> >     verifications: {{ parameters.verifications }}
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > `vmr-template.yml`
> > ```
> > parameters:
> > - name: verifications
> >   type: object
> >   default:
> >     source-build
> >     unified-build
> >     ...etc...
> > 
> > jobs:
> > # Job for android - unified-build
> > - ${{ if and(containsValue(parameters.verifications, 'unified-build'), containsValue(parameters.verifications, 'android')) }}:
> >   - template: ../jobs/vmr-build.yml
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > `vmr-template.yml` definitely needs to have a default list of values, to allow VMR to build all default verification jobs.
> 
> You'd check for the value. I've used this pattern a few times in the VMR. The outer pipeline has a parameter for signing, say, with a string "Default to whatever the default for this branch is". Then the variables template checks for that value and re-interprets it as the appropriate value.

I've seen that and it might work, we'd use some magic value of this parameter for reinterpretation, i.e. `default`.

The alternative is to compress the 3 levels of templates I called above, and only have 2, repo-specific template, that directly references the VMR template.

Then we could pass the `verifications` parameter, if necessary. If not passed, the default set of verifications is used. It would look like this:

`repo-specific.yml - custom verifications`
```
- template vmr-template.yml@vmr
  parameters:
    verifications: source-build
    <other-parameters>
```

`repo-specific.yml - default verifications`
```
- template vmr-template.yml@vmr
  parameters:
    <other-parameters>
```

`vmr-template.yml`
```
parameters:
- name: verifications
  type: object
  default:
    source-build
    unified-build
    ...etc...

jobs:
# Job for android - unified-build
- ${{ if and(containsValue(parameters.verifications, 'unified-build'), containsValue(parameters.verifications, 'android')) }}:
  - template: ../jobs/vmr-build.yml
```

If we used a `default` magic value, and add a variable to the VMR template, we could always have the `verifications` parameter and pass the real value or `default`.

MichaelSimons:
> Here are the final changes that work: https://github.com/dotnet/dotnet/compare/main...vmrverification.list

This looks good.  One thing that I just realized is that we should split the source-build verification into stage1 and stage2.  SBRP should only need stage1 which is what I would expect other repos would want as well.

NikolaMilosavljevic:
> > Here are the final changes that work: [dotnet/dotnet@main...vmrverification.list](https://github.com/dotnet/dotnet/compare/main...vmrverification.list)
> 
> This looks good. One thing that I just realized is that we should split the source-build verification into stage1 and stage2. SBRP should only need stage1 which is what I would expect other repos would want as well.

Yeah, makes sense, I was thinking of that. How about `source-build-stage1` and `source-build-stage2`? Also, if just stage2 job is specified, we would still build stage1 job, as it is a prerequisite.

MichaelSimons:
> Yeah, makes sense, I was thinking of that. How about source-build-stage1 and source-build-stage2? Also, if just stage2 job is specified, we would still build stage1 job, as it is a prerequisite.

Makes sense.

eng/common/templates/vmr-build-pr.yml
```diff
@@ -25,9 +16,11 @@ resources:
...
     name: dotnet/dotnet
     endpoint: dotnet
+    ref: refs/heads/vmrverification.list
 
 stages:
 - template: /eng/pipelines/templates/stages/vmr-build.yml@vmr
   parameters:
     isBuiltFromVmr: false
+    verifications: [ "unified-build-browser-wasm" ]
     scope: lite
```
mmitche, line 26: WHat's the difference between `scope` and `verifications`?

PR #1332, opened 2025-07-28
Remove abstractions-xunit

MichaelSimons:
This is no longer being referenced, as surfaced with the [new SBRP usage report](https://github.com/dotnet/source-build/issues/3405).  It should have been [removed with xunit](https://github.com/dotnet/source-build-reference-packages/pull/1288).  xunit and abstractions-xunit were [added at the same time](https://github.com/dotnet/source-build-externals/pull/167).


mthalman (commented):
There are Xunit-related versions in `eng/Versions.props` that should be cleaned up too.

PR #1343, opened 2025-08-13
IdentityModel CodeQL suppression

mthalman:
This will suppress the CodeQL violations that are showing up for the `azure-activedirectory-identitymodel-extensions-for-dotnet` repo. That team has given the ok that these are safe.

Related: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/3297

### DIFF
--- a/CodeQL.yml
+++ b/CodeQL.yml
@@ -1,0 +1,3 @@
+path_classifiers:
+  library:
+    - "src/externalPackages/src/azure-activedirectory-identitymodel-extensions-for-dotnet/**"


MichaelSimons:
Has the identity model team given indication that they will address https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/3297?  If not it feels wrong to suppress all codeql indefinitely.  Can you open a tracking issue to remove this once they implement the appropriate suppressions?

mthalman:
> If not it feels wrong to suppress all codeql indefinitely.

What is the alternative then?

mthalman:
Exactly. So either way, I think this change is necessary.

MichaelSimons (approved):
Please open a tracking issue for the follow-up.

PR #1527, opened 2026-01-06
Regenerate 10.0 refpack IL with latest ildasm

ericstj:
The IL was missing support for `allows ref struct` generic constraint

Fix this by using latest ildasm to generate il.

I used latest rather than 10.0.0 GA to get this fix -- https://github.com/dotnet/runtime/issues/121686

Still I had to patch the files by reverting the removal of AssemblyMetadataAttribute "source" "source-build-reference-packages". I suspect this is added by some post processing.

I also reverted changes to `// Image base:` comment due to https://github.com/dotnet/runtime/issues/122912

### DIFF
--- a/src/targetPacks/ILsrc/microsoft.netcore.app.ref/10.0.0/ref/net10.0/System.Net.WebHeaderCollection.il
+++ b/src/targetPacks/ILsrc/microsoft.netcore.app.ref/10.0.0/ref/net10.0/System.Net.WebHeaderCollection.il
@@ -354,7 +354,6 @@
 .subsystem 0x0003       // WINDOWS_CUI
 .corflags 0x00000009    //  ILONLY
 // Image base: 0x00007FE82078A000
-// Image base: 0x0x7ff33f424000
 
 
 // =============== CLASS MEMBERS DECLARATION ===================


ericstj:
Running this build here, but I plan to port this to https://github.com/dotnet/dotnet/pull/3873

ericstj:
@MichaelSimons @NikolaMilosavljevic -- you may want to regenerate all the ref-packs with a newer ILASM.  Ideally add documentation + scripts for that process since we'll need to do it annually. cc @ViktorHofer

src/targetPacks/ILsrc/microsoft.netcore.app.ref/10.0.0/ref/net10.0/System.Net.WebHeaderCollection.il
```diff
@@ -354,6 +354,7 @@
 .subsystem 0x0003       // WINDOWS_CUI
 .corflags 0x00000009    //  ILONLY
 // Image base: 0x00007FE82078A000
+// Image base: 0x0x7ff33f424000
```
MichaelSimons: Was this intentional?

```suggestion
```

Issue #1313, opened 2025-03-20
[SBRP] Use GenAPI to generate the targeting packs


MichaelSimons:
Currently the targeting packs are based on IL and utilize IlAsm/IlDasm for building/generating the source. In theory we could utilize GenAPI to generate the targeting packs into c# code.

There are unknowns if GenAPI is mature enough to handle shared framework assemblies.  A POC would be needed to prove this out.

APICompat should help validate this change.

The benefits of this would be:

1. Removing the IlAsm/IlDasm dependency.  This adds additional bootstrap packages to the source build prep bootstrap process. e.g. this would help simplify the platform onboard process. This also unifies the build process between ref/targeting packs.  Currently the targeting pack build infra is tied to Linux (see https://github.com/dotnet/source-build/issues/3329)
3. Consolidate the generate tooling to be based on one technology (e.g. GenAPI).  Currently the targeting pack generator is not even in the public repo (see https://github.com/dotnet/source-build/issues/3091)

PR #1528, opened 2026-01-06
Add generate tooling support for target packs

MichaelSimons:
This PR ports the [internal tooling](https://dev.azure.com/dnceng/internal/_git/dotnet-source-build-utilities?path=/IlGenerator/README.md) for generating targeting packs and integrates it with the existing generate tooling.

This is intended to be a lost cost solution for getting the tooling in the open so that devs can self-service adding/updating target packs.  The long term desire is to [utilize GenAPI](https://github.com/dotnet/source-build-reference-packages/issues/1313).

1. All of ugliness with the internal tooling around retrieving the nuget package was removed because the existing tooling provides all of this infra.
2. The new tasks and GenerateTargetPackageSource Target are a port from the existing code.
3. [ILTooling.targets](https://github.com/dotnet/source-build-reference-packages/pull/1528/changes#diff-2f02fb08cfc4f095373bd0e1c798a2ebe1a7cee8e8ceb1db3972815c045e1c70) was factored out of the existing logic for building target packs so that it can be shared with the generate tooling.

Note: There are a couple ildasm tooling bugs that cause non-reproducibility and differences across platforms that are currently handled via the tests.  Regenerating with the latest tooling will solve a couple of these but I would like to handle that with a separate PR.

### DIFF
--- a/eng/ILTooling.targets
+++ b/eng/ILTooling.targets
@@ -2,33 +2,9 @@
   <!-- Shared targets for acquiring and resolving ILAsm/ILDasm tooling. -->
 
   <Target Name="GetILToolPackageReferences">
-    <!-- Detect if we're running on musl-based Linux (Alpine, etc.) -->
-    <Exec IgnoreExitCode="true" Command="ldd --version 2&gt;&amp;1 | grep -q musl" Condition="$([MSBuild]::IsOSPlatform('linux'))">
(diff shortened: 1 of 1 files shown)


eng/ILTooling.targets
ViktorHofer: I think this could all be done without any targets involved - during evaluation but doesn't matter.

ViktorHofer (approved):

PR #1585, opened 2026-03-02
Add external antlr4 project

jkoritzinsky:
We are adding a dependency on Antlr4.Runtime.Standard in https://github.com/dotnet/runtime/pull/124764 for our new managed ilasm implementation. This PR adds the requisite logic to source-build just the Antlr4 C# runtime.

See https://github.com/dotnet/runtime/pull/124764 for more info

cc: @am11

### DIFF
--- a/src/externalPackages/projects/antlr4.proj
+++ b/src/externalPackages/projects/antlr4.proj
@@ -8,6 +8,6 @@
 
     <DotNetToolArgs>$(DotNetToolArgs) /p:Configuration=$(Configuration)</DotNetToolArgs>
 
-    <BuildCommand>$(DotNetTool) build -c $(Configuration) -f netstandard2.0 $(Antlr4ProjectPath) /bl:$(ArtifactsLogRepoDir)build.binlog</BuildCommand>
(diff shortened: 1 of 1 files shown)


src/externalPackages/projects/antlr4.proj
```diff
@@ -0,0 +1,13 @@
+<Project Sdk="Microsoft.Build.NoTargets">
+
+  <PropertyGroup>
+    <!-- Project Paths -->
+    <Antlr4Directory>$(ProjectDirectory)runtime/CSharp/src/</Antlr4Directory>
```
akoeplinger: do we just need this directory? we could cloak the rest of the repo via https://github.com/dotnet/dotnet/blob/main/src/source-mappings.json#L92-L100 for the VMR so that we don't get Cpp/Go/Java/JavaScript/Swift sources to avoid them being picked up by 1ES scanners

ellahathaway (commented):
```
NU1101: Unable to find package Microsoft.NETFramework.ReferenceAssemblies`
```

On Windows, `Antlr4.csproj` targets `net45;netstandard2.0`; restoring `net45` implicitly requires `Microsoft.NETFramework.ReferenceAssemblies`, but this build uses a generated NuGet config that clears all feeds except `SBRP`, where that package is not available.

am11:
> restoring `net45` implicitly requires `Microsoft.NETFramework.ReferenceAssemblies`, but this build uses a generated NuGet config that clears all feeds except `SBRP`, where that package is not available.

I think we should skip net45 as the consumer project is only targeting net11.

PR #1594, opened 2026-03-09
Build antlr4 nupkg

am11:
Needed by https://github.com/dotnet/dotnet/pull/5262.

```sh
$ ./build.sh -sb
$ find artifacts/ -name *.nupkg | grep -i antlr4
..
artifacts/packages/Release/Shipping/Antlr4.Runtime.Standard.4.13.1.nupkg
```


akoeplinger:
Thanks. just FYI you could've done the change in your VMR PR directly too.

PR #1621, opened 2026-04-01
Onboard OpenTelemetry libraries to source-build

baronfel:
This uses the documentation in this repo to have Copilot onboard several OpenTelemetry libraries to the external packages at their latest versions.  This would allow SDK/MSBuild to take dependencies on them for telemetry exports.

We care about the following libraries for this effort:

# Packages that expose OTel APIs
* OpenTelemetry.Api
* OpenTelemetry.Api.ProviderBuilderExtensions
* OpenTelemetry

# Packages that provide exporters for OTel data
* OpenTelemetry.Exporter.OpenTelemetryProtocol

# Packages that gather and report data from applications at runtime
* OpenTelemetry.Instrumentation.Http
* OpenTelemetry.Instrumentation.Runtime

I was able to verify these build with `./build.sh -sb` in a WSL instance.

### DIFF
--- a/tests/SbrpTests/GenerateScriptTests.cs
+++ b/tests/SbrpTests/GenerateScriptTests.cs
@@ -26,6 +26,7 @@ public class GenerateScriptTests
     };
 
     public string SandboxDirectory { get; set; }
+    public string PkgsSandboxDirectory { get; set; }
     public ITestOutputHelper Output { get; set; }
 
     public GenerateScriptTests(ITestOutputHelper output)
@@ -33,6 +34,8 @@ public class GenerateScriptTests
         Output = output;
         SandboxDirectory = Path.Combine(Environment.CurrentDirectory, $"GenerateTests-{DateTimeOffset.UtcNow.ToUnixTimeSeconds().ToString()}");
         Directory.CreateDirectory(SandboxDirectory);
+        PkgsSandboxDirectory = Path.Combine(SandboxDirectory, ".packages");
+        Directory.CreateDirectory(PkgsSandboxDirectory);
     }
 
     [Theory]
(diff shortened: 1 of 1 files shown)


src/externalPackages/patches/opentelemetry-dotnet-contrib/0001-Source-build-disable-CPM-use-BundledNETCoreAppTarget.patch
```diff
@@ -0,0 +1,301 @@
+From 06b66ff7a60813baf4553c4e867b58c5c8ffba7c Mon Sep 17 00:00:00 2001
```
MichaelSimons: A general question: have you considered if any of the patch changes are worthy of pushing upstream?  Changes like making UseArtifactsOutput overridable is an example as it would allow us to specify a build time property versus requiring a patch to Directory.Build.props.

src/externalPackages/patches/opentelemetry-dotnet/0001-Source-build-disable-CPM-use-BundledNETCoreAppTarget.patch
```diff
@@ -0,0 +1,246 @@
...
+-  -->
+-  <ItemGroup>
+-    <!-- Default to latest versions for all TFMs -->
+-    <PackageVersion Include="Microsoft.Extensions.DependencyInjection.Abstractions" Version="[10.0.0,)" />
+-    <PackageVersion Include="Microsoft.Extensions.Diagnostics.Abstractions" Version="[10.0.0,)" />
+-    <PackageVersion Include="Microsoft.Extensions.Hosting.Abstractions" Version="[10.0.0,)" />
+-    <PackageVersion Include="Microsoft.Extensions.Logging.Configuration" Version="[10.0.0,)" />
+-  </ItemGroup>
+-  <!-- Version overrides for specific netX.0 TFMs -->
+-  <ItemGroup Condition="'$(TargetFramework)' == 'net8.0'">
```
MichaelSimons: Is it necessary to remove these ItemGroups for the removed TFMs?  Keeping the patches to a minimum helps reduce patch maintenance burden.

MichaelSimons:
> The Windows failure is because of path too long errors on a specific file during cloning - the impacted file is in tests, which aren't relevant for this repo, so are we able to mask the submodule repos that we clone?

There isn't an existing mechanism within SBRP.  We can cloak at the VMR level if desired.  We should arguable support long paths regardless within CI.

MichaelSimons:
@baronfel - The revised patches are great.  They will be much easier to maintain.  Thanks for enhancing the patch maintainability guidelines/instructions.  Looks like there is a restore issue.  When that is resolved I will signoff.

MichaelSimons:
> I will log a follow-up issue on this.

https://github.com/dotnet/source-build-reference-packages/issues/1623

MichaelSimons (approved):

PR #1638, opened 2026-04-09
Update Newtonsoft.Json to 13.0.4

mthalman:
Updates the Newtonsoft.Json external package from 13.0.3 to 13.0.4.

Changes:
- Updated submodule to 13.0.4 tag
- Updated version in Versions.props
- Regenerated patch for 13.0.4 csproj
- Updated SourceRevisionId and FileVersionRevision in project file

Related to https://github.com/dotnet/dotnet/pull/5791#issuecomment-4216462688


ellahathaway (commented):
Is it okay to make this change before https://github.com/dotnet/dotnet/pull/5791 is merged? Would doing so result in a 13.0.3 prebuilt until the Arcade flow change lands? If so, should we roll this into https://github.com/dotnet/dotnet/pull/5791 instead?

mthalman:
> Is it okay to make this change before [dotnet/dotnet#5791](https://github.com/dotnet/dotnet/pull/5791) is merged? Would doing so result in a 13.0.3 prebuilt until the Arcade flow change lands? If so, should we roll this into [dotnet/dotnet#5791](https://github.com/dotnet/dotnet/pull/5791) instead?

Yes, it'll cause a prebuilt. I'll need to roll the two code flows together. I need to make the change here first because it's changing a submodule. You can't make such a change directly in the VMR because the VMR has the raw sources files, not the submodule link, so backflow wouldn't work.

ellahathaway (approved):

PR #1667, opened 2026-05-18
Replace cleanup-unreferenced-packages pipeline with AI skill

MichaelSimons:
## Summary

Replaces the existing `cleanup-unreferenced-packages` Azure DevOps pipeline with an AI skill that can be invoked interactively to intelligently clean up unreferenced packages.

## Motivation

The existing pipeline had several failure modes that made it impractical to use without manual intervention:

1. **Code flow timing** — Packages added to SBA tip-of-main that haven't yet flowed into the VMR would be incorrectly flagged as unreferenced. Running the pipeline would delete them, requiring manual re-addition.

2. **Pending uptake** — Packages are sometimes added to SBA ahead of the consuming repo merging the PR that depends on them. These appear unreferenced in the VMR build but are legitimately needed.

3. **Dependency graphs** — Packages that are dependencies of pending-uptake packages also get flagged, creating a cascade of false positives.

Implementing business logic to handle all these scenarios in a static pipeline would be complex and brittle. An AI skill is much better suited because it can:
- Diff SBA against what's flowed to the VMR via `source-manifest.json`
- Prompt the user about known pending-uptake work
- Resolve transitive dependency graphs on-demand
- Present results for human confirmation before acting

## Changes

- **Added**: `.github/skills/cleanup-unreferenced-packages/SKILL.md` — AI skill runbook with 7-step workflow
- **Added**: `.github/skills/cleanup-unreferenced-packages/cleanup-packages-config.json` — Known false positives config
- **Deleted**: `azure-pipelines/builds/cleanup-unreferenced-packages.yml` — Old pipeline
- **Updated**: `README.md` — Reflects new skill-based workflow, fixes `/azp run` command references

Cleanup PR created using this skill - https://github.com/dotnet/source-build-assets/pull/1666


mthalman (commented):
It looks like everything here is capable of being completely automated except for the user escape hatch to keep certain packages from being deleted that have uptake still in progress. I'm trying to weigh that against the toil that this introduces, requiring a dev to run this skill on a regular basis.

Is there a way we can make this completely automated with no human involvement? What if we had a more formal process for new packages to be added that the skill could query that would help answer that question from above without requiring human input?

MichaelSimons:
> I'm trying to weigh that against the toil that this introduces, requiring a dev to run this skill on a regular basis.

My envision is to utilize a monthly scheduled workflow, but yes it would require human involvement to approve the packages to delete.

> Is there a way we can make this completely automated with no human involvement? What if we had a more formal process for new packages to be added that the skill could query that would help answer that question from above without requiring human input?

Two ideas come to mind:

1. Require issues - The idea is the skill can track back to the PR that added the deleted package and see if the tracking issue is still open.  Personally I see this as adding more cumulative toil.  Today tracking issues are not used regularly.  Requiring them would be more toil.  Not enforcing them would cause toil when packages are deleted before consumed.
2. Loop in the author - The idea is the skill would delete all unused packages and open a PR.  If it detects a deleted package was recently added, it would loop in the package contributor to get confirmation the package is not necessary.   The message would indicate how to get copilot to add the package back if needed.

MichaelSimons:
Talking this over with @mthalman , we like the idea of looping in the author as part of the long-term approach here.

The thinking is that we would first check whether an unreferenced package was recently added and apply a grace period before taking any action. For example, if the package was added and the cleanup task runs within two weeks, we would simply leave it alone.

Then, for packages that fall into a middle window — say between two weeks and two months old — we would notify the author who added the package and ask whether it is expected to be consumed soon. The notification could explain that the package is currently being flagged as unreferenced and, if it is still work in progress, provide instructions (potentially Copilot-driven) for restoring or re-adding it as appropriate.

Finally, once a package falls outside that grace window, the skill would simply delete it. We'd want the grace period to be sufficiently large to avoid disrupting active work.

I think that's the longer-term direction we want to experiment with and evaluate over time.

That said, Matt and I both agreed that the current skill is already a significant improvement over the existing pipeline, which is not providing any value today. Because of that, we're going to proceed with merging this PR, and I'll log a follow-up issue to track the additional work needed to make this fully autonomous and not require human interaction.

At the moment this is a lower-priority improvement area, and I have some other work items that need attention first.

mthalman (approved):

MichaelSimons:
> I'll log a follow-up issue to track the additional work needed to make this fully autonomous and not require human interaction.

https://github.com/dotnet/source-build-assets/issues/1670

PR #1668, opened 2026-05-18
Add package validation test for API compatibility with upstream

### LINKED ISSUE #1424: Run PackageValidation to ensure that produced packages are public API identical with upstream versions
APICompat's PackageValidation compares generated packages against baseline versions. SBRP is a perfect use-case for that to ensure that produced packages are public API identical to the upstream versions. PackageValidation should be configured to run in strict mode to enforce no additions and no omissions.

This would need to be conditioned on whether online sources are available to not introduce new prebuilts.

cc @NikolaMilosavljevic @MichaelSimons

MichaelSimons:
Adds a unit test that validates all reference packages and targeting packs produced by this repo are public-API compatible with their upstream NuGet counterparts using apicompat's PackageValidation in strict mode.

## Key design decisions

- Uses `Microsoft.DotNet.ApiCompat.Tool` in package mode with strict baseline validation (both additions and omissions are flagged)
- Tool version managed via `Version.Details.xml` (flows from VMR automatically)
- Invoked via `dotnet exec` (`PackageDownload` + resolved DLL path — no `dotnet-tools.json` needed)
- TFM exclusions shared between the generator and test via `eng/SourceBuildExcludedFrameworks.props` (single source of truth)
- Only packages with source under `src/referencePackages/src` or `src/targetPacks/ILsrc` are validated
- PKV004/PKV006 suppressed (expected for compile-time-only packages that have no runtime assets)
- Intra-package TFM consistency check disabled (`--run-api-compat false`) because polyfill packages intentionally differ across TFMs
- Download failures are test errors (not silently skipped)
- Packages processed in parallel for throughput

## Files changed

| File | Purpose |
|------|---------|
| `eng/SourceBuildExcludedFrameworks.props` | Shared TFM exclusion list used by both generator and test |
| `eng/Version.Details.xml` | Added `Microsoft.DotNet.ApiCompat.Tool` dependency |
| `eng/Version.Details.props` | Auto-generated version property for the tool |
| `eng/Versions.props` | (no new entry — version flows from Version.Details) |
| `tests/SbrpTests/PackageCompatibilityTests.cs` | The new test |
| `tests/SbrpTests/Sbrp.Tests.csproj` | PackageDownload + runtime config for tool path and exclusions |
| `src/packageSourceGenerator/PackageSourceGenerator.proj` | Now imports shared exclusion list |

Fixes #1424

### DIFF
--- a/src/targetPacks/Directory.Build.targets
+++ b/src/targetPacks/Directory.Build.targets
@@ -46,7 +46,7 @@
   <Target Name="BuildIlFile" DependsOnTargets="ResolveIlToolPaths">
     <Error Condition="'$(IlasmPath)' == ''" Text="The 'IlasmPath' property must be set." />
 
-    <Exec Command="$(IlasmPath) $(IlFile) -dll -quiet -nologo &quot;-output=$(OutputFile)&quot;"
+    <Exec Command="$(IlasmPath) $(IlFile) -dll -quiet -nologo -deterministic &quot;-output=$(OutputFile)&quot;"
           IgnoreStandardErrorWarningFormat="true"
           CustomErrorRegularExpression=": error : " />
   </Target>


ViktorHofer:
Curious, why did you decide to use the ApiCompat.Tool package and run this as part of a test? Today, this shipping documented feature automatically runs after Pack as part of an msbuild task. The perf with that approach is much better and required inputs are handled automatically.

The Tool package only exists for outside of msbuild use cases.

MichaelSimons:
Copilot closed this on my behalf because it messed up all of the changes....

> Curious, why did you decide to use the ApiCompat.Tool package and run this as part of a test?

Lack of knowledge in this area in addition to copilot leading me down the path of using the tool.  I am investigating `EnablePackageValidation` on the projects.

PR #1671, opened 2026-05-20
Enable package validation with baseline comparison

### LINKED ISSUE #1424: Run PackageValidation to ensure that produced packages are public API identical with upstream versions

MichaelSimons:
Enables package validation (API compat) for reference packages and targeting packs, comparing the built packages against their upstream NuGet baseline to ensure public API surface compatibility.

Fixes #1424

## What this does

- Enables `EnablePackageValidation` with baseline comparison in `eng/PackageValidation.targets`
- Adds `eng/ExcludeTargetFrameworks.props` as the single source of truth for TFMs that source-build doesn't include, shared by both the package source generator and package validation
- Configures `PackageValidationBaselineFrameworkToIgnore` to skip baseline comparisons for excluded TFMs, eliminating false-positive CP0001 errors for TFMs we intentionally don't build
- Enables strict mode for baseline validation so real API breaks (type removals, renames) are caught
- Adds `CompatibilitySuppressions.xml` files for packages where the generator produces known differences from the baseline
- Updates documentation (README and copilot instructions) with the API compat workflow

## Suppressions

| Diagnostic | Count | Affected packages |
|-----------|-------|-------------------|
| CP0021 — Missing `notnull` generic constraint | 612 | system.collections.immutable, system.text.json, microsoft.extensions.* |
| CP0008 — Missing interfaces | 25 | microsoft.build, microsoft.build.utilities.core |
| CP0002 — Missing event accessor members | 4 | microsoft.build |
| CP0001 — Missing types | 2 | system.configuration.configurationmanager (needs investigation) |


mthalman (approved):

eng/ExcludeTargetFrameworks.props
```diff
@@ -0,0 +1,15 @@
...
+  <!-- Target frameworks that source-build doesn't support/include.
+       This list is consumed by both the package source generator (for filtering which TFMs to
+       generate) and package validation (for ignoring baseline comparisons against absent TFMs).
+       Keep the two uses in sync by defining the canonical list here. -->
+  <PropertyGroup>
+    <ExcludeTargetFrameworks>monoandroid*;monotouch*;net20;net35;net4*;net5.0;netcore50;netcoreapp*;netstandard1*;portable*;uap*;win8;win81;wp8;wpa81;xamarin*</ExcludeTargetFrameworks>
+
+    <!-- Same list with '*' escaped as '%2A' for use in MSBuild ItemGroup Include attributes where
+         wildcards would otherwise be treated as file globs. -->
+    <ExcludeTargetFrameworksMSBuildEscaped>monoandroid%2A;monotouch%2A;net20;net35;net4%2A;net5.0;netcore50;netcoreapp%2A;netstandard1%2A;portable%2A;uap%2A;win8;win81;wp8;wpa81;xamarin%2A</ExcludeTargetFrameworksMSBuildEscaped>
```
ViktorHofer: I would avoid this duplication. Consider using MSBuild's escaping infra: https://learn.microsoft.com/en-us/visualstudio/msbuild/how-to-escape-special-characters-in-msbuild?view=visualstudio#msbuild-special-characters

eng/PackageValidation.targets
```diff
@@ -0,0 +1,20 @@
+<Project>
+
+  <!-- Package validation: compare built packages against their upstream NuGet baseline to ensure
+       API surface compatibility. Disabled when building from the VMR since upstream feeds aren't
+       available in that environment. -->
+  <PropertyGroup Condition="'$(DotNetBuildFromVMR)' != 'true' and '$(IsPackable)' != 'false' and '$(PackageVersion)' != ''">
```
ViktorHofer: NuGet itself conditions on IsPackable=true. I would that here as well.

ViktorHofer: The PackageVersion condition is a bit problematic IMO. It would silently no validate packages that don't have correctly configured package versions. Was that condition necessary?

src/referencePackages/src/microsoft.build.utilities.core/17.11.48/CompatibilitySuppressions.xml
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- https://learn.microsoft.com/dotnet/fundamentals/package-validation/diagnostic-ids -->
+<Suppressions xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:xsd="http://www.w3.org/2001/XMLSchema">
+  <Suppression>
```
ViktorHofer, line 5: Even though xml comments are not preserved currently when re-generating these files, we advise to add them here so that it's clear what these suppressions are about.

src/referencePackages/Directory.Build.targets
```diff
@@ -1,8 +1,28 @@
...
+         Source-generated reference packages use simplified version metadata that doesn't match upstream. -->
+    <!-- PKV006: "TFM is no longer supported in the latest version."
+         Missing TFMs are expected here because of TFM filtering. Source-build itself
+         validates that the required frameworks are present. -->
+    <NoWarn>$(NoWarn);CP0003;PKV006</NoWarn>
+  </PropertyGroup>
+
+  <!-- Ignore baseline TFMs that source-build doesn't support/include. Uses the same exclusion list as the
+       package source generator. -->
+  <ItemGroup Condition="'$(EnablePackageValidation)' == 'true'">
```
ViktorHofer: nit: I would remove this condition given that you don't have it on the above imports or the above NoWarn.

ViktorHofer (commented):
This looks good. Do you understand the compatibility differences? LLM might be able to summarize them. I'm curious about anything that stands out that would indicate a misconfiguration in this repo or something that GenAPI should do differently.

eng/ExcludeTargetFrameworks.props
MichaelSimons: I can't simply call msbuild's escape method because it escapes the semicolon as well causing the string to be treated as one when passed to api compat.  To eliminate the duplication between the two lists, I updated the PackageSourceGenerator to use an ItemGroup just like api compat for the TFMs.  https://github.com/dotnet/source-build-assets/pull/1671/changes/11c450114bf5e22b46b197a1c1ffd6e606629f15

eng/PackageValidation.targets
MichaelSimons: fixed in https://github.com/dotnet/source-build-assets/pull/1671/changes/11c450114bf5e22b46b197a1c1ffd6e606629f15

MichaelSimons: No I don't think this is necessary.  Removed in https://github.com/dotnet/source-build-assets/pull/1671/changes/11c450114bf5e22b46b197a1c1ffd6e606629f15

src/referencePackages/Directory.Build.targets
MichaelSimons: Removed in https://github.com/dotnet/source-build-assets/pull/1671/changes/11c450114bf5e22b46b197a1c1ffd6e606629f15

src/referencePackages/src/microsoft.build.utilities.core/17.11.48/CompatibilitySuppressions.xml
MichaelSimons: Addressed in https://github.com/dotnet/source-build-assets/pull/1671/changes/113007fc7281dc36fe3d6b9c04443ef5b8daaf0d.

Noting https://github.com/dotnet/sdk/issues/45928 as the underlying tracking issue to improve the UX here.

src/packageSourceGenerator/PackageSourceGeneratorTask/GetPackageItems.cs
```diff
@@ -39,12 +39,12 @@ public partial class GetPackageItems : Task
         /// <summary>
         /// Include only the package compile items and dependencies with a matching target framework.
         /// </summary>
-        public string? IncludeTargetFrameworks { get; set; }
+        public ITaskItem[]? IncludeTargetFrameworks { get; set; }
```
ViktorHofer: nit: this could also just be a `string[]?` for simplicity. Same below.

ViktorHofer (approved):

src/packageSourceGenerator/PackageSourceGeneratorTask/GetPackageItems.cs
MichaelSimons: fixed in https://github.com/dotnet/source-build-assets/pull/1671/changes/6dc9cb48a61e03bf03535c1e98d27118a8bc2c25

MichaelSimons:
> This looks good. Do you understand the compatibility differences? LLM might be able to summarize them. I'm curious about anything that stands out that would indicate a misconfiguration in this repo or something that GenAPI should do differently.

From my investigation compliments of LLM, I logged the following issues:

* https://github.com/dotnet/sdk/issues/54450
* https://github.com/dotnet/sdk/issues/54451
* https://github.com/dotnet/sdk/issues/54464

ViktorHofer (approved):

MichaelSimons:
FYI, I logged one more genapi issue - https://github.com/dotnet/sdk/issues/54464

PR #1672, opened 2026-05-20
Add patched 8.0 system.security.cryptography packages

wtgodbe:
So they can be used in aspnetcore (8.0.0 crypto.xml gets a CG alert). Will need to backport this to the servicing branches as well.


MichaelSimons:
@wtgodbe, can you point to the aspnetcore PR that is requiring this?  I'm curious if this is a transitive dependency for you coming through msbuild?

wtgodbe:
Yep, root dependency is `Microsoft.Build.Tasks.Core`. PR is https://github.com/dotnet/aspnetcore/pull/66765

PR #1707, opened 2026-06-04
Align OpenTelemetry source-build with shipped NuGet versions

### LINKED ISSUE #5577: OpenTelemetry source-build packages have wrong AssemblyVersion and missing experimental APIs
## Summary

After the SDK upgraded OpenTelemetry to 1.15.3 in https://github.com/dotnet/dotnet/commit/1616fce8866, source-built OTel packages diverge from Microsoft-built ones:

1. **Wrong AssemblyVersion** — MSFT: `1.0.0.0`, SB: `1.15.3.0`
2. **Missing experimental APIs** — Types gated by `EXPOSE_EXPERIMENTAL_FEATURES` are absent
3. **Assembly identity mismatch** — Contrib packages differ in 4th version component (CI BuildNumber)

See diffs in https://github.com/dotnet/dotnet/pull/6892/changes/6ebbf9039fa4e3ca91c6c16e657f73d8617d3bcd

## Root Cause

`opentelemetry-dotnet.proj` and `opentelemetry-dotnet-contrib.proj` in source-build-assets pass `/p:IntegrationBuild=true` which disables MinVer. Without MinVer, the target chain that sets `AssemblyVersion` and enables experimental features never fires.

mthalman:
## Summary

Source-built OpenTelemetry packages diverged from the Microsoft-shipped NuGet packages on `AssemblyVersion`, `FileVersion`, and the `ExposeExperimentalFeatures`-gated API surface. This PR fixes the divergence entirely in source-build-assets (mirroring prior version-override fixes like #1631, #1610, #1630), and along the way extends the existing validation infrastructure so this class of drift is caught by tests rather than discovered downstream.

Fixes https://github.com/dotnet/source-build/issues/5577.

---

## Background — the version-drift problem

Many external packages we re-build use [MinVer](https://github.com/adamralph/minver) (or a similar git-tag-driven scheme) to compute `AssemblyVersion`, `FileVersion`, and `InformationalVersion` from git tags + CI run numbers. Source-build sets `/p:IntegrationBuild=true`, which removes the `MinVer` `PackageReference` in upstream builds. The MinVer-driven target that sets the version properties then never runs, so source-built assemblies fall back to defaults that do not match the shipped NuGet packages.

For OpenTelemetry specifically:
- `AssemblyVersion` fell back to the assembly default instead of MinVer's stable `Major.0.0.0`.
- `FileVersion` equaled the plain release version (e.g. `1.15.3`) instead of `1.15.3.<runNumber>`.
- `ExposeExperimentalFeatures` stayed at its default `true`, exposing API (e.g. `OpenTelemetry.Logs.Logger`) that the shipped 1.15.3 packages mark `internal`.

The pattern of overriding these values in `.proj` files in this repo is established (see the prior version-fix PRs above). What was missing was a way to *validate* that the overrides we hand-author actually match what Microsoft shipped — and that they stay correct when submodules bump.

---

## What this PR does

Three layers, each in its own commit:

### 1. The OpenTelemetry fix (commit `3cce751e`)

`opentelemetry-dotnet.proj`:
- `AssemblyVersionOverride=1.0.0.0` (MinVer's stable `Major.0.0.0` for the core packages)
- `FileVersionRevision=2017` (the GitHub Actions run number for the `core-1.15.3` release; same across all five core packages)
- `ExposeExperimentalFeatures=false` on the build
- Opt in to validation by listing the package as a `FileVersionValidationPackage`

`opentelemetry-dotnet-contrib.proj`:
- Per-package `FileVersionRevision` values (Http=`1025`, Runtime=`1029`) — each contrib package is released independently from its own package-specific git tag and carries its own CI run number.
- `AssemblyVersion` = `FileVersion` = `ReleaseVersion.BuildNumber` for both, matching the upstream contrib MinVer target.

### 2. Validation: from property to item + MSBuild-evaluated metadata (commits `1e9a4746`, `4136987b`)

The repo already had `FileVersionValidationPackage` (a property) plus the `UpdateExternalMetadata` tool and `ExternalPackageTests`. They covered exactly one package per component and one validated aspect (`FileVersion` revision).

This PR generalizes the surface so a single `.proj` can declare *multiple* independently-versioned packages and validate `AssemblyVersion`, `FileVersion`, and `InformationalVersion` independently:

- **Item form, not property form.** `FileVersionValidationPackage` is now an MSBuild item. Per-item `*Property` metadata names the MSBuild property that holds each override value. `null`/missing means the item does not opt in to that aspect.

  ```xml
  <ItemGroup>
    <FileVersionValidationPackage Include="OpenTelemetry.Instrumentation.Http"
                                  FileVersionRevisionProperty="HttpFileVersionRevision"
                                  AssemblyVersionOverrideProperty="HttpAssemblyVersionOverride" />
    <FileVersionValidationPackage Include="OpenTelemetry.Instrumentation.Runtime"
                                  FileVersionRevisionProperty="RuntimeFileVersionRevision"
                                  AssemblyVersionOverrideProperty="RuntimeAssemblyVersionOverride" />
  </ItemGroup>
  ```

- **MSBuild evaluation (not XDocument parsing).** Commit `1e9a4746` originally read items via XDocument. Commit `4136987b` switched to loading the file as a true `Microsoft.Build` `Project`, which unlocked `<ItemDefinitionGroup>` defaults — conventional property names (e.g. `FileVersionRevisionProperty=FileVersionRevision`) so single-item `.proj`s don't need to repeat them:

  ```xml
  <ItemGroup>
    <FileVersionValidationPackage Include="Newtonsoft.Json" />
  </ItemGroup>
  ```

- **Explicit vs. defaulted bindings, distinguished.** An item that *explicitly* declares it opts into an aspect (e.g. `AssemblyVersionOverrideProperty="Foo"`) is a hard error if `Foo` is undefined — the author's intent is clear. An item that picks up an aspect *only* via an `ItemDefinitionGroup` default and finds the property undefined silently skips that aspect — the author never explicitly opted in.

- **Other safety checks added in `4136987b`:**
  - Metadata values containing literal `$(…)`, `@(…)`, or `%(…)` are rejected at parse time using the *un-evaluated* metadata value. This catches authors who write `<FileVersionRevisionProperty>$(Foo)</…>` thinking metadata is a property reference (it isn't — it's a literal property *name*), and catches undefined-expression cases even when the referenced property doesn't exist.
  - Two items resolving to the same target property → collision error (avoids silent last-write-wins).

### 3. `UpdateExternalMetadata` automation

The CLI auto-rewrites override values from package metadata so authors don't have to hand-edit them when a submodule bumps:

```
dotnet UpdateExternalMetadata.dll opentelemetry-dotnet
```

It updates `SourceRevisionId` in the `.proj` and (for components with sibling `.props`) rewrites the validated override properties to match the values in the shipped NuGet package, preserving formatting via line-buffer surgical edits rather than `XDocument` serialization round-trips.

---

## Architecture — per-component `.props` files

Initial attempts to define these version properties within each `.proj` file hit a hard wall in the test host:

- vstest preloads `Microsoft.Build, Version=15.1.0.0` before any `[ModuleInitializer]` runs.
- `MSBuildLocator.RegisterDefaults()` cannot run inside `dotnet test` once that has happened.
- Any file declaring `Sdk="Microsoft.Build.NoTargets"` requires SDK resolution, which requires the locator.

The escape: validation metadata lives in **SDK-less, standalone `.props` files**, one per component, sibling to the component's `.proj`:

```
src/externalPackages/projects/
├── validation.props                   ← shared ItemDefinitionGroup defaults only
├── newtonsoft-json.proj               ← imports newtonsoft-json.props
├── newtonsoft-json.props              ← per-component overrides + items
├── opentelemetry-dotnet.proj
├── opentelemetry-dotnet.props
├── opentelemetry-dotnet-contrib.proj
├── opentelemetry-dotnet-contrib.props
└── …
```

These `.props` files declare no `Sdk` and have no upward import chain, so they load with plain `new Project(path)` in any host — no `MSBuildLocator` dependency anywhere on the test host path. The shared `validation.props` holds only the `<ItemDefinitionGroup>` defaults; each `<component>.props` imports it.

`SourceRevisionId` stays in the `.proj` (it's a build-time MSBuild property, not validation metadata). The `UpdateExternalMetadata` tool updates `SourceRevisionId` unconditionally, and only enters the validation-metadata phase when the component has a sibling `.props`. This means unmigrated components (e.g. `humanizer`) still benefit from the `SourceRevisionId` automation.


ellahathaway (commented):
My main concern with this is that we'll have to update the packages with each revision, which could be frequent? If that's a valid concern, is there a way to pin the package references [here](https://github.com/dotnet/sdk/blob/22c3477bb8247b03a33ea553ba1460814ed7823d/eng/Versions.props#L63-L64) to specific revisions?

mthalman:
> My main concern with this is that we'll have to update the packages with each revision, which could be frequent? If that's a valid concern, is there a way to pin the package references [here](https://github.com/dotnet/sdk/blob/22c3477bb8247b03a33ea553ba1460814ed7823d/eng/Versions.props#L63-L64) to specific revisions?

That is the intent of https://github.com/dotnet/source-build-assets/blob/main/eng/update-external-metadata.sh, to automate the version updating when updating the submodule. And that's verified by the tests in https://github.com/dotnet/source-build-assets/blob/main/tests/SbrpTests/ExternalPackageTests.cs.

I still have more work to do to wire things up for these changes.

tests/SbrpTests/ExternalPackageTests.cs
```diff
@@ -84,75 +92,82 @@ public void SourceRevisionIdMatchesSubmoduleCommit()
...
     }
 
+    private static string? ReadSourceRevisionIdFromProj(string projFile)
+    {
+        XDocument doc = XDocument.Load(projFile);
+        return doc.Root?
+            .Elements("PropertyGroup")
+            .Elements("SourceRevisionId")
+            .FirstOrDefault()
+            ?.Value;
```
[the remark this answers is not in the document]
mthalman: Fixed in c8d227ce

tests/SbrpTests/ExternalPackageTests.cs
```diff
@@ -84,75 +92,82 @@ public void SourceRevisionIdMatchesSubmoduleCommit()
...
     public void ReleaseVersionMatchesPackageOutput()
     {
         string artifactsObjDir = Path.Combine(RepoRoot, "artifacts", "obj");
+        Project versionsProject = CommonUtilities.LoadProject(VersionsPropsPath);
 
         List<string> errors = new();
         int checkedCount = 0;
 
-        foreach ((string projFile, string packageName, XDocument doc) in LoadProjectFiles())
+        foreach ((string projFile, string packageName, Project project) in LoadProjectFiles())
```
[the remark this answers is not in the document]
mthalman: Fixed in 39263f63

tests/SbrpTests/ExternalPackageTests.cs
```diff
@@ -167,111 +182,190 @@ public void ReleaseVersionMatchesPackageOutput()
...
     {
+        Project versionsProject = CommonUtilities.LoadProject(VersionsPropsPath);
         List<string> errors = new();
         int checkedCount = 0;
 
-        foreach ((string projFile, string packageName, XDocument doc) in LoadProjectFiles())
+        // Cache per (packageId, version) so multiple aspects/items don't re-download.
+        Dictionary<string, PackageVersionMetadata> metadataCache = new();
+
+        foreach ((string projFile, string packageName, Project project) in LoadProjectFiles())
```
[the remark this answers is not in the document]
mthalman: Fixed in 68f7b109

src/externalPackages/projects/opentelemetry-dotnet-contrib.props
ellahathaway, line 1: **Coverage gap — contrib `AssemblyVersion` is build-pinned but never validated** (Confidence: Medium)

The `.proj` passes `/p:AssemblyVersion=$(HttpFileVersion)` (i.e., `AssemblyVersion = FileVersion` = `1.15.1.1025`), but this `.props` only binds `FileVersionRevisionProperty`. The default `AssemblyVersionOverrideProperty` (from `validation.props`) targets `AssemblyVersionOverride`, which this file doesn't define — so the aspect is silently skipped ("validate if present" behavior).

Result: `AssemblyVersionOverrideMatchesPublishedPackage` never covers contrib packages. If upstream contrib uses MinVer's `Major.0.0.0` (like core OTel), the build may ship an incorrect `AssemblyVersion` without the validation tests catching it.

Consider either (a) adding an explicit `AssemblyVersionOverrideProperty` + corresponding property per contrib package, or (b) documenting why contrib intentionally differs from core's `AssemblyVersion=1.0.0.0` pattern.

eng/tools/UpdateExternalMetadata/Program.cs
ellahathaway, line 1: **Bug — silent data corruption via `$` in Regex replacement** (Confidence: High)

In `UpdatePropertyLine`, the `replacement` string is passed directly as the replacement pattern to `Regex.Replace`:

```csharp
string replacement = $"<{elementName}>{new XText(newValue)}</{elementName}>";
lines[lineIndex] = Regex.Replace(originalLine, pattern, replacement);
```

`Regex.Replace` interprets `$1`, `$&`, `${name}`, `$$` etc. as substitution tokens in the replacement string. `new XText(newValue)` escapes XML chars (`&`, `<`, `>`) but **not** `$`. If any version string ever contains `$` (e.g., an `InformationalVersion` with SemVer build metadata), the written value will be silently corrupted.

Fix: use the `MatchEvaluator` overload to bypass substitution parsing:
```csharp
lines[lineIndex] = Regex.Replace(originalLine, pattern, _ => replacement);
```

ellahathaway, line 1: **Defensive gap — no validation that `newValue` is single-line** (Confidence: Medium)

In `UpdatePropertyLine`, if `newValue` ever contains `\n` or `\r`, it would inject a newline into a single buffer entry, causing the buffer's line count to drift from the `IXmlLineInfo` line numbers captured during the initial XDocument parse. All subsequent `UpdatePropertyLine` calls on the same buffer would then target wrong lines.

The multi-update-per-file correctness relies on the invariant that each edit is single-line (no line count change). Version strings shouldn't contain newlines, but there's no guard.

Consider adding: `Debug.Assert(!newValue.Contains('\n') && !newValue.Contains('\r'));` or throwing if violated.

eng/tools/CommonUtilities/CommonUtilities.cs
ellahathaway, line 1: **Resource leak — `ProjectCollection` never disposed** (Confidence: High)

```csharp
public static Project LoadProject(string filePath)
{
    ProjectCollection collection = new();
    return new Project(filePath, globalProperties: null, toolsVersion: null, collection);
}
```

`ProjectCollection` implements `IDisposable` and owns evaluation caches, a logger registry, and toolset state. The caller can never dispose it because the reference is dropped on return. This is invoked repeatedly in tests (`LoadProjectFiles()` per component × 4 test methods) and the updater tool, accumulating undisposed collections for the process lifetime.

Consider returning a disposable wrapper that exposes both the `Project` and its owning `ProjectCollection`, or have callers manage the collection lifetime via `using`.

tests/SbrpTests/ExternalPackageTests.cs
ellahathaway, line 1: **Test behavior change — `ReleaseVersionMatchesPackageOutput` silently skips instead of failing** (Confidence: Medium)

In the refactored no-items branch of `ReleaseVersionMatchesPackageOutput`:

```csharp
if (items.Count == 0)
{
    string? releaseVersion = CommonUtilities.FindReleaseVersion(versionsProject, packageName);
    if (string.IsNullOrEmpty(releaseVersion))
    {
        continue;   // checkedCount NOT incremented
    }
    ...
}
```

The old code (`FindReleaseVersionOrError`) added an error to the list when the release version was not found. The refactored code silently continues. If all non-items components hit this path and no project has items, `Skip.If(checkedCount == 0)` will skip the entire test — masking a bug where release versions were renamed/removed from `eng/Versions.props`.

ellahathaway (commented):
Review of the validation infrastructure generalization and OpenTelemetry fix. Found 6 issues total:

**Posted inline (4):**
- 1 bug (Regex `$` metachar in replacement — silent data corruption)
- 1 resource leak (ProjectCollection never disposed)
- 1 coverage gap (contrib AssemblyVersion unvalidated)
- 1 test behavior change (silent skip instead of fail)

**Could not post inline (2):**
- **FileEditBuffer.cs** — Mixed line endings: `content.Contains("\r\n")` picks one newline style for `Split`, but `IXmlLineInfo` counts both. With mixed endings, buffer indices drift from XML line numbers, causing wrong-line edits. Fix: split on `"\r\n"` and `"\n"` simultaneously.
- **Program.cs (UpdatePropertyLine)** — No validation that `newValue` is single-line. If it contained `\n`, the buffer line count would drift from IXmlLineInfo expectations, corrupting subsequent edits. Consider asserting single-line.

eng/tools/UpdateExternalMetadata/Program.cs
mthalman: Fixed in 36fd6ee0

src/externalPackages/projects/opentelemetry-dotnet-contrib.props
mthalman: Fixed in 1d39d30a

eng/tools/UpdateExternalMetadata/Program.cs
mthalman: Fixed in 06da53e9

tests/SbrpTests/ExternalPackageTests.cs
mthalman: There is no behavior change here. See https://github.com/dotnet/source-build-assets/pull/1707/changes#diff-f122352b238d8ad8ea154d7ec3b5453eacc537a38f3f5d45d21e3a3e204a4b31L148-L153

eng/tools/CommonUtilities/CommonUtilities.cs
mthalman: Fixed in 762ae284

tests/SbrpTests/ExternalPackageTests.cs
```diff
@@ -167,111 +183,195 @@ public void ReleaseVersionMatchesPackageOutput()
...
+
+                string? releaseVersion = ResolveReleaseVersion(versionsProject, item, packageName);
+                if (string.IsNullOrEmpty(releaseVersion))
+                {
+                    errors.Add($"{packageName}.proj: {overrideAspectName} item for package '{item.PackageId}': " +
+                        $"could not resolve release version " +
+                        $"({(item.ReleaseVersionPropertyName is null ? "auto-derive failed" : $"property '{item.ReleaseVersionPropertyName}' not in eng/Versions.props")}).");
+                    checkedCount++;
+                    continue;
+                }
```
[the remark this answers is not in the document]
mthalman: Fixed in b8b613f6

tests/SbrpTests/ExternalPackageTests.cs
```diff
@@ -167,111 +183,195 @@ public void ReleaseVersionMatchesPackageOutput()
...
             }
 
-            checkedCount++;
+            foreach (ValidationPackageItem item in items)
+            {
+                string? releaseVersion = ResolveReleaseVersion(versionsProject, item, packageName);
+                if (string.IsNullOrEmpty(releaseVersion))
+                {
+                    errors.Add($"{packageName}.proj: Could not resolve release version for validation package '{item.PackageId}' " +
+                        $"({(item.ReleaseVersionPropertyName is null ? "auto-derived" : $"explicit ReleaseVersionProperty='{item.ReleaseVersionPropertyName}'")}).");
```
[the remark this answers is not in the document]
mthalman: Fixed in 3fac3fb0

tests/SbrpTests/ExternalPackageTests.cs
```diff
@@ -167,111 +183,195 @@ public void ReleaseVersionMatchesPackageOutput()
...
+                        $"download/extraction is broken — verify the published package contents.");
+                    checkedCount++;
+                    continue;
+                }
+
+                if (expectedValue != actualValue)
+                {
+                    errors.Add($"{packageName}.proj: {binding.PropertyName} '{expectedValue}' does not match " +
+                        $"actual '{actualValue}' from {item.PackageId} {releaseVersion}.");
+                }
```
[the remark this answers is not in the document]
mthalman: Fixed in dba14231

ellahathaway (approved):

PR #1754, opened 2026-07-30
Add reference packages required by roslyn

akhera99:
Add reference packages required by the Roslyn dependency update in dotnet/dotnet#7891.
 
 The update introduces package versions that are not currently available in source-build-assets, causing the dotnet/dotnet source-only build to report prebuilt package usage.
 
 ## Packages added
 
 - Microsoft.Build 18.7.1
 - Microsoft.Build.Framework 18.7.1
 - Microsoft.Build.Tasks.Core 18.7.1
 - Microsoft.Build.Utilities.Core 18.7.1
 - Microsoft.NET.StringTools 18.7.1
 - System.CodeDom 10.0.8
 - System.Formats.Nrbf 10.0.4
 - System.Reflection.MetadataLoadContext 10.0.4
 - System.Security.Cryptography.Pkcs 10.0.8
 - System.Security.Cryptography.Xml 10.0.8
 - Required transitive reference packages
 
 Existing customizations were carried forward for the new package versions, and API compatibility suppressions were generated for known GenAPI limitations.
 
 ## Validation
 
 Built, packed, and ran API compatibility validation for all 23 generated reference-package projects.
 
 Fixes the prebuilt detection failure in https://github.com/dotnet/dotnet/pull/7891.


mthalman (commented):
This is adding a package version with a vulnerability. Need to upgrade it.

```
 warning NU1903: Package 'System.Security.Cryptography.Xml' 10.0.8 has a known high severity vulnerability, https://github.com/advisories/GHSA-23rf-6693-g89p
```

Also, it's fine to iterate in this PR. But don't merge it here. Since https://github.com/dotnet/dotnet/pull/7891 is dependent on it, I would recommend just applying these changes in that PR. Otherwise, you'll likely end up with the dependency flow of these changes also failing because they require the changes in https://github.com/dotnet/dotnet/pull/7891. So the easiest way is to bundle them together from the start.

